PR #1, opened 2015-08-06
25503 v2

jfboismenu-adsk:
Creating this for code review before Alan goes.


app.py
```diff
@@ -9,126 +9,74 @@
...
+class MultiWorkFiles(sgtk.platform.Application):
 
     def init_app(self):
         """
         Called as the application is being initialized
         """
+        self._tk_multi_workfiles = self.import_module("tk_multi_workfiles")
 
         application_has_scenes = True
         if self.engine.name == "tk-mari":
```
jfboismenu-adsk: Someday we'll have self.engine.has_scene(), right? :)

app.py
```diff
@@ -9,126 +9,74 @@
...
+            self.engine.register_command("File Save...", self.show_file_save_dlg, {"short_name":"file_save"})
+
+        # Process auto startup options - but only on certain supported platforms
         # because of the way QT inits and connects to different host applications
         # differently, in conjunction with the 'boot' process in different tools,
         # the behaviour can be very different.  
-        
+
         # currently, we have done basic QA on nuke and maya so we limit these options to 
         # those two engines for now.
```
jfboismenu-adsk: Not your comment, but while we're there: those three engines

PR #2, opened 2015-10-13
For #30809, list view icon is visible but not implemented.

jfboismenu-adsk:
- Remove view switching buttons since only one view type was enabled.
- A bit more files we're touched by this branch because I'm using a different version of PySide. I figured I'd left them in so that in the future we don't always have to remove them.


jfboismenu-adsk:
The best code submissions are the ones where there is no new code. :)

josh-t:
Looks good! Good to merge. Back to IP.

PR #3, opened 2015-10-22
For #31249, overwriting scene files

jfboismenu:
If we let the model refresh it's state before clicking save, clicking it will invoke _generate_path and the cached value will be read.

If we don't wait for the background task to finish indexing the files, then we will get partial results and an incorrect version number.

This is because when the FileModel requests for the cache file versions it doesn't honour the clean_only flag that _generate_path to it. Because of this, the file model actually returns dirty cache entries. Honouring the flag when communicating with the cache will return None when the model is still being populated and the _generate_path will compute the value itself on the spot instead of waiting for the background task.

### DIFF
--- a/python/tk_multi_workfiles/file_model.py
+++ b/python/tk_multi_workfiles/file_model.py
@@ -377,7 +377,7 @@ class FileModel(QtGui.QStandardItemModel):
                             searching for files.
         :returns:           A dictionary {version:FileItem} of all file versions found.
         """
-        return self._search_cache.find_file_versions(work_area, key)
+        return self._search_cache.find_file_versions(work_area, key, include_dirty=not clean_only)
 
     def items_from_file(self, file_item, ignore_version = False):
         """
--- a/python/tk_multi_workfiles/file_model.py
+++ b/python/tk_multi_workfiles/file_model.py
@@ -377,7 +377,7 @@ class FileModel(QtGui.QStandardItemModel):
                             searching for files.
         :returns:           A dictionary {version:FileItem} of all file versions found.
         """
-        return self._search_cache.find_file_versions(work_area, key, include_dirty=not clean_only)
+        return self._search_cache.find_file_versions(work_area, key, clean_only)
 
     def items_from_file(self, file_item, ignore_version = False):
         """
--- a/python/tk_multi_workfiles/file_save_form.py
+++ b/python/tk_multi_workfiles/file_save_form.py
@@ -67,17 +67,17 @@ class FileSaveForm(FileFormBase):
             preview_colour = font_colour.darker(140)
         self._preview_colour = (preview_colour.red(), preview_colour.green(), preview_colour.blue())
 
-        self._initializing = True
+        self._disallow_preview_update = True
         try:
             # doing this inside a try-except to ensure any exceptions raised don't 
             # break the UI and crash the dcc horribly!
             self._do_init()
-            self._initializing = False
+            self._disallow_preview_update = False
             # Manually invoke the preview update here so it is only called once due to the
-            # _initializing flag.
+            # _disallow_preview_update flag.
             self._start_preview_update()
         except:
-            self._initializing = False
+            self._disallow_preview_update = False
             app.log_exception("Unhandled exception during File Save Form construction!")
 
     def _do_init(self):
@@ -195,7 +195,7 @@ class FileSaveForm(FileFormBase):
         # to avoid creating and deleting the preview task multiple times, which is not only
         # wasteful but also makes the code harder to debug since 5 tasks end up computing
         # the preview.
-        if self._initializing:
+        if self._disallow_preview_update:
             return
 
         # Disable the button while the path is computed.
--- a/python/tk_multi_workfiles/file_save_form.py
+++ b/python/tk_multi_workfiles/file_save_form.py
@@ -333,7 +333,7 @@ class FileSaveForm(FileFormBase):
                 # and raise a new, clearer exception for this specific use case:
                 raise TankError("Unable to resolve template fields!  This could mean there is a mismatch "
                                 "between your folder schema and templates.  Please email "
-                                "toolkitsupport@shotgunsoftware.com if you need help fixing this.")
+                                "support@shotgunsoftware.com if you need help fixing this.")
 
             # it's ok not to have a path preview at this point!
             return {}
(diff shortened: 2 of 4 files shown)


jfboismenu-adsk:
While this fixes a bug in the file model, this might not be the ideal workflow. Maybe a better user experience would be to gray out the Save button while the model is computing the correct file name. Then when the name is ready we can update un-gray the button.

python/tk_multi_workfiles/file_model.py
```diff
@@ -377,7 +377,7 @@ def get_cached_file_versions(self, key, work_area, clean_only=False):
                             searching for files.
         :returns:           A dictionary {version:FileItem} of all file versions found.
         """
-        return self._search_cache.find_file_versions(work_area, key)
+        return self._search_cache.find_file_versions(work_area, key, include_dirty=not clean_only)
```
manneohrstrom: it looks a little bit funny that this method has a `clean_only` param only to call another method which has a `include_dirty` parameter. All that it's doing is to negate the flag and call the other method. Makes me wonder if these methods could be combined into a single one? I don't have the proper big picture so may not make sense at all, but i thought it's worth at least checking!

python/tk_multi_workfiles/file_save_form.py
```diff
@@ -68,13 +68,20 @@ def __init__(self, parent=None):
             preview_colour = font_colour.darker(140)
         self._preview_colour = (preview_colour.red(), preview_colour.green(), preview_colour.blue())
 
+        self._initializing = False
         try:
             # doing this inside a try-except to ensure any exceptions raised don't 
             # break the UI and crash the dcc horribly!
+            self._initializing = True
```
manneohrstrom: Is this needed? Looks like this would be equivalent:

```
self._initializing = True
try:
   do_stuff()
except:
   self._initializing = False
```

or maybe even

```
self._initializing = True
try:
   self._do_init()
except:
   app.log_exception("Unhandled exception during File Save Form construction!")
finally:
   self._initializing = False

self._start_preview_update()
```

python/tk_multi_workfiles/file_save_form.py
```diff
@@ -182,6 +189,14 @@ def _on_use_next_available_version_toggled(self, checked):
     def _start_preview_update(self):
         """
         """
+        # When initializing the gui, the event is fired multiple times due to signals
```
manneohrstrom: Maybe this could be moved up into the empty doc string?

manneohrstrom:
Some minor style comments on the code. I cannot comment (yet) on the overall structure and approach - would be great to get @alandann's take on that! 

The approach with the gray button sounds great to me. Sounds like it could potentially simplify the code too - as a consequence of not having to handle the "loading state" - we just know that no click events will come in during that phase if we disable the button...

python/tk_multi_workfiles/file_save_form.py
jfboismenu: We can always move the first assignment outside the if, but the second example you provided isn't equivalent. If something goes wrong during start_preview_update in your example, the exception will bubble out of the __init__.

python/tk_multi_workfiles/file_model.py
alandann: ignoring the dirty/clean misname (I agree, I know I wrote it but it should be consistent!) this is indeed the piece that was missing :(  This was added specifically to ensure that the file save didn't decide on an incorrect version to save to if it managed for example to find all publishes before it found all work files.

python/tk_multi_workfiles/file_save_form.py
```diff
@@ -68,13 +67,19 @@ def __init__(self, parent=None):
             preview_colour = font_colour.darker(140)
         self._preview_colour = (preview_colour.red(), preview_colour.green(), preview_colour.blue())
 
+        self._initializing = True
```
alandann: I did try to get this to behave correctly by ordering everything inside do_init but the fact it wasn't is a good indication that was a fragile approach.  One thing I would possibly say is that naming it more specifically might make more sense - e.g. call it self._allow_preview_update or something as it's only used for that purpose.

python/tk_multi_workfiles/file_save_form.py
```diff
@@ -181,7 +186,21 @@ def _on_use_next_available_version_toggled(self, checked):
...
+        # When initializing the gui, events are fired multiple times due to signals
+        # emitted from the widgets. Here we are muting the start_preview_update call
+        # to avoid creating and deleting the preview task multiple times, which is not only
+        # wasteful but also makes the code harder to debug since 5 tasks end up computing
+        # the preview.
+        if self._initializing:
+            return
+
+        # Disable the button while the path is computed.
+        self._disable_save("Please wait while Toolkit calculates the next available file name.")
```
alandann: I'm not sure this is needed as generate_path is called again before the save operation (https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/3/files#diff-1d0c9979f5f3e3247a4b3cb32ee3445aR664) so even if the user hits save straight away before the preview has been generated everything should be ok.  The main problem with greying out the button is that it tends to flicker as users click around and feels a bit rubbish!

It's up to you guys but personally, as long as the preview generation is fixed (the include_dirty=not clean_only above) then I'd leave this out.

The only other thing to make it even more robust would be to perform a complete search (rather than using the cached result) when the user hits save.  

There is also an additional check in the save logic to see if the file exists before it's actually overwritten - currently that the user can choose to overwrite the file but it could be changed to just return to this dialog and force them to try again...

jfboismenu: > The only other thing to make it even more robust would be to perform a complete search (rather than using the cached result) when the user hits save.
Actually we are already not using the cached result by not asking for the dirty results (clean_only=True) in the _generate_path method. Or is there something else we're missing?

jfboismenu:
@manneohrstrom 
> IMO, having state tracking like that self._initializing tends to add maintenance cost and overhead and is often an indication that we are working around something else. I really think we should consider the "disabled button" approach here instead - from what i understand, this seems simpler and cleaner.

Actually, the _allow_preview_update variable is meant to reduce the amount of noise since even widget that gets updated fire an update event which relaunches preview generation. This has nothing to do with the gray button and is all about making the output of the logger more easily debuggable.

PR #6, opened 2015-12-07
For #30812, update path cache when opening file manager

jfboismenu:
Now updating the path cache when opening the file manager or refreshing the views.

### DIFF
--- a/python/tk_multi_workfiles/file_form_base.py
+++ b/python/tk_multi_workfiles/file_form_base.py
@@ -231,6 +231,10 @@ class FileFormBase(QtGui.QWidget):
 
         :param checked:    True if the refresh action is checked - ignored
         """
+        app = sgtk.platform.current_bundle()
+        app.log_debug("Synchronizing remote path cache...")
+        app.sgtk.synchronize_filesystem_structure()
+        app.log_debug("Path cache up to date!")
         self._refresh_all_async()
 
     def _refresh_all_async(self):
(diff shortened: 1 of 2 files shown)


python/tk_multi_workfiles/file_form_base.py
```diff
@@ -231,6 +231,10 @@ def _on_refresh_triggered(self, checked=False):
 
         :param checked:    True if the refresh action is checked - ignored
         """
+        app = sgtk.platform.current_bundle()
```
jfboismenu: This will sync the path cache every single time the file manager (open and save) is opened.

python/tk_multi_workfiles/work_files.py
```diff
@@ -125,10 +125,21 @@ def wrapper(*args, **kwargs):
...
 class WorkFiles(object):
     """
     Main entry point for all commands in the app.
     """
+
+    def __init__(self):
+        """
+        Constructor.
+        """
+        app = sgtk.platform.current_bundle()
```
jfboismenu: This will sync the path cache every single time the file manager (open and save) is opened.

python/tk_multi_workfiles/file_form_base.py
jfboismenu: This will sync the path cache every single time the user hit's F5 or Command-R/Ctrl-R

PR #7, opened 2015-12-10
For #33041, previous versions menu

jfboismenu:
- Allows the user to open previous versions of the current selection in the file browser. The items shown in the sub-menu are the same that you would have gotten if you had clicked directly on that specific version in the "All versions" mode.
- Added documentation around the action system.
- These files were mostly PEP-8, I couldn't resist fixing them for good. :)

### DIFF
--- a/python/tk_multi_workfiles/actions/action.py
+++ b/python/tk_multi_workfiles/actions/action.py
@@ -1,19 +1,18 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
 """
 
 import sgtk
-from sgtk import TankError
-from sgtk.platform.qt import QtGui, QtCore
+
 
 class Action(object):
     """
@@ -23,15 +22,29 @@ class Action(object):
         """
         self._app = sgtk.platform.current_bundle()
         self._label = label
-        
+
     @property
     def label(self):
         return self._label
-    
+
     def execute(self, parent_ui):
         """
         """
         raise NotImplementedError("Implementation of _execute() method missing for action '%s'" % self.label)
+
+
+class ActionGroup(Action):
+
+    def __init__(self, label, actions):
+        """
+        """
+        Action.__init__(self, label)
+        self.__actions = actions
+
+    @property
+    def actions(self):
+        return self.__actions
+
 
 class SeparatorAction(Action):
     """
--- a/python/tk_multi_workfiles/actions/action.py
+++ b/python/tk_multi_workfiles/actions/action.py
@@ -9,49 +9,77 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
+Menu actions.
 """
 
 import sgtk
 
 
-class Action(object):
+class ActionBase(object):
     """
+    Base class for Actions.
     """
     def __init__(self, label):
         """
+        Constructor.
+
+        :param label: Name of the action.
         """
         self._app = sgtk.platform.current_bundle()
         self._label = label
 
     @property
     def label(self):
+        """
+        :returns: The name of the action.
+        """
         return self._label
+
+
+class Action(ActionBase):
+    """
+    Base class for leaf actions, ie, actions that when selected execute a piece of logic. This logic
+    is implemented in the execute method.
+    """
 
     def execute(self, parent_ui):
         """
+        Called when the user executes a given action. The default implementation raises a NotImplementedError.
+
+        :raises NotImplementedError: Thrown if a derived class doesn't implement this method and the client invokes it.
         """
         raise NotImplementedError("Implementation of _execute() method missing for action '%s'" % self.label)
 
 
-class ActionGroup(Action):
+class ActionGroup(ActionBase):
+    """
+    Group of actions.
+    """
 
     def __init__(self, label, actions):
         """
+        Constructor.
+
+        :param label: Name of the group of actions.
+        :param actions: List of ActionBase actions.
         """
-        Action.__init__(self, label)
-        self.__actions = actions
+        ActionBase.__init__(self, label)
+        self.__actions = actions[:]
 
     @property
     def actions(self):
+        """
+        :returns: List of ActionBase actions.
+        """
         return self.__actions
 
 
-class SeparatorAction(Action):
+class SeparatorAction(ActionBase):
     """
+    Not an actual action but a hint to the UI that a separation should be shown.
     """
     def __init__(self):
-        Action.__init__(self, "---")
-
-    def execute(self, parent_ui):
-        # do nothing!
-        pass
+        """
+        Constructor.
+        """
+        ActionBase.__init__(self, "---")
--- a/python/tk_multi_workfiles/actions/file_action_factory.py
+++ b/python/tk_multi_workfiles/actions/file_action_factory.py
@@ -125,7 +125,7 @@ class FileActionFactory(object):
 
         actions.extend(self._create_local_file_actions(self._file_item))
         actions.extend(self._create_published_file_actions(self._file_item))
-        actions.extend(self._create_previous_versions_actions(self._file_item))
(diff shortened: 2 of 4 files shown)


python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,12 +37,39 @@
...
+        :param workfiles_visible: True if workfiles are visible.
+        :param publishes_visible: True if publishes are visible.
+
+        :returns: List of Actions.
+        """
+        return self.__get_actions(file, work_area, file_model, workfiles_visible, publishes_visible, first_level=True)
+
+    def __get_actions(self, file, work_area, file_model, workfiles_visible, publishes_visible, first_level):
+        """
+        Retrives the list of actions for the given file, recursively.
```
josh-t: Maybe there's a better way to organize this method? It says it's recursive, but it's seemingly a fixed, single recursion to handle the previous version submenus. It's already a long method, so maybe that's another reason to break it up a bit and just have individual methods for returning the different types of actions rather than one big one with recursion. Just an idea. 

*extremely* simplified example:

```
def __get_actions(...):
  
    # ...
    actions.extend(self.__local_file_actions(...))
    actions.extend(self.__published_file_actions(...))
    actions.extend(self.__first_level_actions(...))
    actions.extend(self.__custom_actions(...))

    return actions

def __first_level_actions(...):

    # ...
    fl_actions.extend(self.__local_file_actions(...))
    fl_actions.extend(self.__published_file_actions(...))
    fl_actions.extedn(self.__custom_actions(...))
    
    return [group_action(fl_actions)]
```

Hope that makes sense.

python/tk_multi_workfiles/actions/action.py
```diff
@@ -1,44 +1,85 @@
...
+
+        :param label: Name of the action.
         """
         self._app = sgtk.platform.current_bundle()
         self._label = label
-        
+
     @property
     def label(self):
+        """
```
josh-t: Is the `label` used outside of the `Action` class? Do the `SeparatorAction` or `ActionGroup` classes use the label? Just wondering if it would make sense to move it into the `Action` class.

josh-t: Thinking about it, I guess the `ActionGroup` label makes sense.

python/tk_multi_workfiles/actions/action.py
```diff
@@ -1,44 +1,85 @@
...
+Menu actions.
 """
 
 import sgtk
-from sgtk import TankError
-from sgtk.platform.qt import QtGui, QtCore
 
-class Action(object):
+
+class ActionBase(object):
```
josh-t: Just wondering if this might be better named `MenuItemBase` or something similar?

python/tk_multi_workfiles/actions/file_action_factory.py
jfboismenu: Excellent point, will do. I might as well fix that. Thanks for bringing that up.

python/tk_multi_workfiles/actions/action.py
jfboismenu: I'd avoid the term Menu item because it then implies we have a menu, which is implementation detail of the workfiles dialog.

jfboismenu: The fact that we generate separators however does violate that rule tough ;)

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+            self._can_copy_to_work_area = current_env.work_template is not None
             # (AD) TODO - it's possible the work template for the current work area has different requirements than
-            # the source work area (e.g. it may require a name where the source work template doesn't!)  This should 
+            # the source work area (e.g. it may require a name where the source work template doesn't!)  This should
             # probably check that file.path is translatable to the current work area (contains all the required keys)
             # in addition to the simple check it's currently doing.
 
-        # get the list of file versions for the file:
-        file_versions = file.versions
+    def get_actions(self, file_item):
```
josh-t: needs documentation of `file_item` param.

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+            return []
+
+        actions = []
+        # ------------------------------------------------------------------
+        actions.append(SeparatorAction())
+        actions.append(NewFileAction(self._work_area))
+
+        return actions
+
+    def _create_custom_actions(self, file_item, file_versions):
```
josh-t: need docs for `file_versions`

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+
+            previous_versions_actions.append(ActionGroup("Version %d" % previous_file_item.version, version_actions))
+
+        return previous_versions_actions
+
+    def _create_new_file_action(self):
+        """
+        Creates a list of actions to create a new file.
+
+        :param file_item: File item to generate actions for.
```
josh-t: Looks like this arg was removed.

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+        version_versions_actions = self._create_previous_versions_actions(
+            previous_versions
+        )
+        return [ActionGroup(label, version_versions_actions)]
+
+    def _create_previous_versions_actions(self, previous_versions):
+        """
+        Creates a list of actions for all the previous versions of a given type.
+
+        :param pick_locals: True if we want to show previous actions for workfiles, False if we want them for publishes.
```
josh-t: Different name or different arg. Anyway, should be updated to document `previous_versions`.

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+
 class FileActionFactory(object):
     """
+    Builder for the file open menu. Based on the current selection, the build will enumerate a list of actions,
+    grouped actions and seperators that should be displayed in the UI.
     """
-    
-    def get_actions(self, file, work_area, file_model, workfiles_visible=True, publishes_visible=True):
+
+    def __init__(self, work_area, file_model, workfiles_visible=True, publishes_visible=True):
```
josh-t: Now that you're doing part of the work in `__init__`, it would imply that the Factory is less dynamic. Meaning that when the work area changes or the model changes, you have to create a new factory instance. I'm guessing that's fine and potentially gives a slight performance boost? I'm just curious about the change really since I'm not familiar with how the class is used in the app. :)

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+        )
 
         # determine if this file is in a different work area:
-        user_work_area = work_area
-        change_work_area = (work_area.context != app.context)
-        if change_work_area and in_other_users_sandbox:
-            user_work_area = work_area.create_copy_for_user(g_user_cache.current_user)
-            change_work_area = (user_work_area.context != app.context)
+        self._user_work_area = work_area
+        self._change_work_area = (work_area.context != app.context)
```
josh-t: might be nice to have a couple of quick comments in here indicating what `_user_work_area` and `_change_work_area` represent.

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+                self._publishes_visible
+            )
+        )
+
+        actions.extend(self._create_local_file_actions(file_item, current_user_file_versions))
+        actions.extend(self._create_published_file_actions(file_item, current_user_file_versions))
+        actions.extend(self._create_previous_versions_actions_menus(file_item))
+        actions.extend(self._create_new_file_action())
+        actions.extend(self._create_custom_actions(file_item, current_user_file_versions))
+        actions.extend(self._create_show_in_actions(file_item, current_user_file_versions))
```
josh-t: wondering if this would be a good place to just give a quick overview of what each of the groups of actions do or what they're used for. maybe with examples. not really necessary, just might help future devs when processing the code.

python/tk_multi_workfiles/actions/file_action_factory.py
```diff
@@ -36,158 +37,336 @@
...
+        Creates a list of actions for all the previous versions of a given type.
+
+        :param pick_locals: True if we want to show previous actions for workfiles, False if we want them for publishes.
+
+        :returns: List of Actions.
+        """
+        previous_versions_actions = []
+
+        # Gives us the last ten versions, from the latest to the earliest.
+        for previous_file_item in previous_versions[-1:-11:-1]:
```
josh-t: maybe make the number of versions to show a variable set at the class level? Would make it slightly easier to change down the road if need be. very minor.

python/tk_multi_workfiles/actions/file_action_factory.py
jfboismenu: The factor was being instantiated every single time we needed it, so the performance should be the same. Moving some of those variables to data members made the code cleaner since it's static state that doesn't need to be passed around to other methods anymore.

PR #9, opened 2016-01-06
For #34142, don't show entity names in treeview.

jfboismenu:
Initializes the cached schema with a background manager so models can pull the data down.

### DIFF
--- a/python/tk_multi_workfiles/file_form_base.py
+++ b/python/tk_multi_workfiles/file_form_base.py
@@ -24,6 +24,8 @@ BackgroundTaskManager = task_manager.BackgroundTaskManager
 shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
 ShotgunEntityModel = shotgun_model.ShotgunEntityModel
 
+shotgun_globals = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_globals")
+
 from .file_model import FileModel
 from .my_tasks.my_tasks_model import MyTasksModel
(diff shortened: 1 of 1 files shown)


python/tk_multi_workfiles/file_form_base.py
```diff
@@ -79,25 +83,29 @@ def closeEvent(self, event):
...
+            self._entity_models = []
+
+            # and shut down the task manager
+            if self._bg_task_manager:
+                self._bg_task_manager.shut_down()
+                self._bg_task_manager.deleteLater()
+                self._bg_task_manager = None
+        finally:
+            # Make sure that we always unregister our manager from the globals.
+            shotgun_globals.unregister_bg_task_manager(self._bg_task_manager)
```
manneohrstrom: I think you should first unregister all links to the task manager, then destroy it, e.g. the `shotgun_globals.unregister_bg_task_manager(self._bg_task_manager)` call should happen before the shutdown. Also, IMHO the try/finally feels like it may be overkill? Maybe just

```python
            # clear up the various data models:
            if self._file_model:
                self._file_model.destroy()
                self._file_model.deleteLater()
                self._file_model = None
            if self._my_tasks_model:
                self._my_tasks_model.destroy()
                self._my_tasks_model.deleteLater()
                self._my_tasks_model = None
            for _, model in self._entity_models:
                model.destroy()
                model.deleteLater()
            self._entity_models = []

            # and shut down the task manager
            if self._bg_task_manager:
                shotgun_globals.unregister_bg_task_manager(self._bg_task_manager)
                self._bg_task_manager.shut_down()
                self._bg_task_manager.deleteLater()
                self._bg_task_manager = None
```

PR #10, opened 2016-01-07
For #30799, workfiles app shouldn't be modal.

jfboismenu:
The dialog is not modal anymore. The only exception is Nuke 6. Due to issues with deallocation of PySide objects in non-main threads in that version of Nuke, the dialog is modal and a timer runs the garbage collection at regular intervals.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,18 +1,19 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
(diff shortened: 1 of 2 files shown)


python/tk_multi_workfiles/work_files.py
```diff
@@ -96,7 +98,7 @@ def wrapper(*args, **kwargs):
         bytes_before = 0
         if sys.platform == "Darwin":
             import resource
-            bytes_before = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss/1024.0/1024.0
+            bytes_before = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024.0 / 1024.0
```
jfboismenu: The next three fixes were all that was missing to silence PEP-8, so I figured I might as well do it.

app.py
```diff
@@ -76,25 +78,73 @@ def show_file_save_dlg(self):
...
+    def use_managed_gc(self):
+        """
+        Flag indicating if the dialogs need managed garbage collection.
+
+        :returns: True if managed garbage collection is required, False otherwise.
+        """
+        return self._is_pyside_unstable()
+
+    @property
+    def use_debug_dialog(self):
```
jfboismenu: I discussed with Rob and pointed out that PySide leak detection was helpful for debugging the dialog so we decided to add a hidden setting for us when we work on the app.

PR #11, opened 2016-01-15
Allows context changes. Uses context changing instead of engine resta…

thebeeland:
…rts when setting its target context.


jfboismenu:
I had to diff this on the command line because GitHub can't seem to diff correctly. :(

Looks good to me. Merge whenever.

PR #12, opened 2016-01-28
For #30799, non modal dialog in all DCCs

jfboismenu:
- Dialogs are not modal anymore to behave better with various DCCs who have issues displaying modal dialogs on top of ours.

- App is a lot more stable, especially in Nuke on OSX and Windows. This has been achieved by making all model related objects children of the dialog instead of relying on deleteLater to delete them. While this has stabilized the app across the board since deleteLater causes crashes in multiple DCCs, it also made the models leak in certain DCCs instead of being freed, which is hardly ideal. However, a memory leak at the moment is better than a complete DCC crash. We will look right away on a leak mitigation strategy. There's a strong smell that this leak might be related to PySide and outside our control, since children QObjects should be released by Qt automatically.

- Removed the Shotgun Model Overlay Widget which never worked with the app and might be a source of instability.

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,18 +1,19 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
-Multi Work Files 2.  
+Multi Work Files 2.
 Provides File Open/Save functionality for Work Files
 """
 import sgtk
+
 
 class MultiWorkFiles(sgtk.platform.Application):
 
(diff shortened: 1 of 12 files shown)


python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -90,12 +90,7 @@ def __init__(self, entity_model, search_label, allow_task_creation, parent):
         self._ui.entity_tree.expanded.connect(self._on_item_expanded)
         self._ui.entity_tree.collapsed.connect(self._on_item_collapsed)
 
-        # create the overlay 'busy' widget that will be displayed when the model is reset:
-        self._overlay_widget = ShotgunModelOverlayWidget(None, self._ui.entity_tree)
```
jfboismenu: Removed because it has never been working with the app and I didn't want to have anything that we didn't fully understand mess around with models since they are so brittle/leak friendly in certain DCCs.

python/tk_multi_workfiles/file_list/user_filter_menu.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, parent):
 
         self._current_user_action = QtGui.QAction("Show My Files", self)
         self._current_user_action.setCheckable(True)
-        toggled_slot = lambda toggled, uid=self._current_user_id: self._on_user_toggled(uid, toggled)
+        toggled_slot = lambda toggled: self._on_user_toggled(self._current_user_id, toggled)
```
jfboismenu: Old version of PySide don't like methods with too many arguments, even if they are default arguments.

app.py
```diff
@@ -78,45 +78,6 @@ def show_file_save_dlg(self):
         """
         self._tk_multi_workfiles.WorkFiles.show_file_save_dlg()
 
-    def _is_pyside_unstable(self):
```
jfboismenu: Previous version of this work we were under the impression certain versions of Nuke were having issues and not other DCCs. Turns out it was all versions and now all are stable, so removing the code.

python/tk_multi_workfiles/work_files.py
```diff
@@ -17,70 +17,6 @@
 from .util import report_non_destroyed_qobjects
 
 
-class TimedGc(QtCore.QObject):
```
jfboismenu: All this timed garbaged collection code doesn't work with non-modal dialog and forcing garbage collection doesn't help reclaimed leaked models unfortunately.

PR #14, opened 2016-02-16
33063 Add user activity metrics calls

josh-t:
The calls log the metrics internally and ignore any errors. This prevents the need to force the app/fw users to update to a new core that supports metrics. When a supported core is updated, the metrics will be logged.

### DIFF
--- a/python/tk_multi_workfiles/actions/interactive_open_action.py
+++ b/python/tk_multi_workfiles/actions/interactive_open_action.py
@@ -236,7 +236,11 @@ class InteractiveOpenAction(OpenFileAction):
                     src_path = work_path
                     work_path = local_path
 
-        self._app.log_metric("Open workfile")
+        try:
+            self._app.log_metric("Open workfile")
+        except:
+            # ignore all errors. ex: using a core that doesn't support metrics
+            pass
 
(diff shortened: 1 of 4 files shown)


python/tk_multi_workfiles/actions/interactive_open_action.py
```diff
@@ -236,6 +236,12 @@ def _open_workfile(self, file, env, parent_ui):
                     src_path = work_path
                     work_path = local_path
 
+        try:
+            self._app.log_metric("Open workfile")
```
jfboismenu: I'm wondering if we should track if the user is opening a file from a sandbox and if so where is the destination. @robblau , how do you feel about this?

robblau: For metrics whether user sandboxes are in use or not would be the interesting thing, not this level of detail (definitely not info on the dest).  I think in the original design that was there as a metric, but a low priority one.  No worries about putting it in here.  On folder creation is where I'd expect that to be gathered anyway.

jfboismenu: When I said destination, I meant whether users were opening the file in the sandbox or copying it to theirs first and then opening. Not the actual file path.

I agree tough that this might be too fine grained. It would probably be better to track this at the pipeline level if we ever wanted to track it.

PR #15, opened 2016-02-19
For #34391, workfiles2 in nuke studio

jfboismenu:
Application settings detection wasn't honouring the engine instance name, which means that it was possible to switch from one engine instance to another by mistake. This caused issues with the default config with Nuke and Nuke Studio.

We're now forcing context change to happen to an engine that has the same instance name, just like in workfiles 1.

I've also removed some code that allowed to test the app by using workfiles v1 settings.

### DIFF
--- a/app.py
+++ b/app.py
@@ -24,16 +24,13 @@ class MultiWorkFiles(sgtk.platform.Application):
         self._tk_multi_workfiles = self.import_module("tk_multi_workfiles")
         self.__is_pyside_unstable = None
 
-        application_has_scenes = True
         if self.engine.name == "tk-mari":
             # Mari doesn't have the concept of a current scene so this app shouldn't
             # provide any commands!
             return
 
-        # register commands:
-        if application_has_scenes:
-            self.engine.register_command("File Open...", self.show_file_open_dlg, {"short_name":"file_open"})
-            self.engine.register_command("File Save...", self.show_file_save_dlg, {"short_name":"file_save"})
+        self.engine.register_command("File Open...", self.show_file_open_dlg, {"short_name":"file_open"})
+        self.engine.register_command("File Save...", self.show_file_save_dlg, {"short_name":"file_save"})
 
         # Process auto startup options - but only on certain supported platforms
         # because of the way QT inits and connects to different host applications
--- a/app.py
+++ b/app.py
@@ -28,7 +28,7 @@ class MultiWorkFiles(sgtk.platform.Application):
         if self.engine.name == "tk-mari":
             # Mari doesn't have the concept of a current scene so this app shouldn't
             # provide any commands!
-            application_has_scenes = False
+            return
 
         # register commands:
         if application_has_scenes:
--- a/python/tk_multi_workfiles/actions/file_action.py
+++ b/python/tk_multi_workfiles/actions/file_action.py
@@ -84,33 +84,6 @@ class FileAction(Action):
         if create_folders:
             FileAction.create_folders(ctx)
 
-
-    @staticmethod
-    def restart_engine(ctx):
-        """
-        Set context to the new context.  This will
-        clear the current scene and restart the
-        current engine with the specified context
-        """
-        app = sgtk.platform.current_bundle()
-        app.log_info("Restarting the engine...")
-        
-        # restart engine:
-        QtGui.QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)
-        try:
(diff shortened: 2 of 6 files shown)


app.py
```diff
@@ -28,7 +28,7 @@ def init_app(self):
         if self.engine.name == "tk-mari":
             # Mari doesn't have the concept of a current scene so this app shouldn't
             # provide any commands!
-            application_has_scenes = False
```
jfboismenu: Setting the flag to false meant that the rest of the method wasn't executed, but it wasn't clear from just setting the flag. Returning clarifies the intent.

python/tk_multi_workfiles/actions/file_action.py
```diff
@@ -84,33 +84,6 @@ def create_folders_if_needed(ctx, template):
         if create_folders:
             FileAction.create_folders(ctx)
 
-
-    @staticmethod
```
jfboismenu: Dead code

python/tk_multi_workfiles/actions/file_action.py
```diff
@@ -124,15 +97,9 @@ def change_context(ctx):
         # restart engine:
         QtGui.QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)
         try:
-            # If there's an engine running, change its context. Otherwise
```
jfboismenu: The engine is always running now, so this is not an issue anymore.

python/tk_multi_workfiles/work_area.py
```diff
@@ -18,34 +18,48 @@
...
         def __init__(self):
+            """
+            Constructor.
+            """
             Threaded.__init__(self)
-            self._cache = {}
+            self._cache = []
 
         @Threaded.exclusive
-        def get(self, engine_name, app_name, context):
```
jfboismenu: We're not taking settings from workfiles v1 anymore, nor are we taking settings from other engine instances, so those values are not necessary for computing a key anymore.

python/tk_multi_workfiles/work_area.py
```diff
@@ -18,34 +18,48 @@
...
             """
+            Retrieve the cached settings for a given context.
+
+            :param context: The context for which we desire settings.
+
+            :returns: The settings dictionary or None
             """
-            settings_by_context = self._cache.get((engine_name, app_name), [])
-            for ctx, settings in settings_by_context:
+            for ctx, settings in self._cache:
```
jfboismenu: This could would be cleaner if contexts were hashable.

python/tk_multi_workfiles/actions/file_action.py
```diff
@@ -124,15 +97,9 @@ def change_context(ctx):
...
-            # start the engine up in the new context.
-            if sgtk.platform.current_engine(): 
-                sgtk.platform.change_context(ctx)
-                return
-
-            sgtk.platform.start_engine(current_engine_name, ctx.sgtk, ctx)
+            sgtk.platform.change_context(ctx)
```
manneohrstrom: we might need a min core constraint in info.yaml for this one.... maybe that's specified already?

python/tk_multi_workfiles/work_area.py
```diff
@@ -264,54 +278,39 @@ def _get_settings_for_context(self, context, templates_to_find, settings_to_find
...
+            return app_settings[0].get("settings")
+
+        # There's more than one instance of that app for the engine instance, so we'll
+        # need to deterministically pick one. We'll pick the one with the same
+        # application instance name as the current app instance.
+        for settings in app_settings:
+            if settings.get("app_instance") == app.instance_name:
+                return settings.get("settings")
+
+        return None
```
manneohrstrom: Would it make more sense at this point to log a warning and then pick the first instance found? So we try to provide the user with something half-sensible but also warn about the state of things?

manneohrstrom:
CR looks sweet!

python/tk_multi_workfiles/actions/file_action.py
jfboismenu: Already there.

python/tk_multi_workfiles/work_area.py
jfboismenu: Rob and I debated this. We felt that returning one at random (the ordering of these in memory is not guaranteed) would mean random behaviour, so we decided to return nothing instead. I'll add the warning tough, great idea.

PR #16, opened 2016-04-20
For #36073, remove useless fields from model

jfboismenu:
Having values that are not used in the model impacts performance since we're caching more data than we actually need, so reading and writing that data takes time, but also when that data changes it causes rebuilds of the tree view which can be quite long is the data set is big.

### DIFF
--- a/python/tk_multi_workfiles/work_area.py
+++ b/python/tk_multi_workfiles/work_area.py
@@ -205,23 +205,23 @@ class WorkArea(object):
         settings_to_find = ["saveas_default_name", "saveas_prefer_version_up", 
                             "version_compare_ignore_fields", "file_extensions"]
         resolved_settings = {}
+        app = sgtk.platform.current_bundle()
         try:
             if self._context:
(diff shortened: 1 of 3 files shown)


python/tk_multi_workfiles/work_area.py
```diff
@@ -205,23 +205,23 @@ def _load_settings(self):
...
                             "version_compare_ignore_fields", "file_extensions"]
         resolved_settings = {}
+        app = sgtk.platform.current_bundle()
         try:
             if self._context:
                 resolved_settings = self._get_settings_for_context(self._context, templates_to_find, settings_to_find)
         except:
-            # (TODO) - propogate problems up - maybe add an is_valid() method?
-            pass
+            app.log_exception("There was an error parsing the settings for context '%s'" % self._context)
```
jfboismenu: This is to help a bit with error reporting in workfiles. Everytime we end up here it's because some typo was introduced inside an environment which prevented us from loading it. This will let the user know what broke. Ideally it would be in the UI, but the console will have to do for now.

manneohrstrom:
CR looks fine!

PR #17, opened 2016-04-21
For #35335, user sandbox behaviour fix

jfboismenu:
The user sandbox button is now displayed as soon as a work area CAN have one and not only when there's one on disk. This makes the user experience consistent and predictable.

### DIFF
--- a/python/tk_multi_workfiles/browser_form.py
+++ b/python/tk_multi_workfiles/browser_form.py
@@ -520,12 +520,9 @@ class BrowserForm(QtGui.QWidget):
         if isinstance(form, MyTasksForm):
             # retrieve the selected task from the form and emit a work-area changed signal:
             task, breadcrumb_trail = form.get_selection()
-            # self._emit_work_area_changed(task, breadcrumb_trail)
             self._on_selected_task_changed(task, breadcrumb_trail)
 
         elif isinstance(form, EntityTreeForm):
             # retrieve the selection from the form and emit a work-area changed signal:
             selection, breadcrumb_trail = form.get_selection()
-            # selected_entity = selection.get("entity") if selection else None
-            # self._emit_work_area_changed(selected_entity, breadcrumb_trail)
             self._on_selected_entity_changed(selection, breadcrumb_trail)
--- a/python/tk_multi_workfiles/browser_form.py
+++ b/python/tk_multi_workfiles/browser_form.py
@@ -28,6 +28,7 @@ from .framework_qtwidgets import Breadcrumb
 from .file_list.file_proxy_model import FileProxyModel
 from .file_filters import FileFilters
 from .util import monitor_qobject_lifetime
+from .work_area import WorkArea
 
 class BrowserForm(QtGui.QWidget):
     """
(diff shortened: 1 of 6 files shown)


python/tk_multi_workfiles/browser_form.py
```diff
@@ -1,11 +1,11 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-#
```
jfboismenu: I've limited whitespaces fixes to lines with comments or empty lines as to not confuse git blame.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -163,37 +165,28 @@ def set_models(self, my_tasks_model, entity_models, file_model):
             self._file_model.set_users(self._file_filters.users)
 
             # add an 'all files' tab:
-            all_files_form = FileListForm(self, "All Files", self._file_filters, show_work_files=True, show_publishes=True)
```
jfboismenu: Refactoring. :)

python/tk_multi_workfiles/browser_form.py
```diff
@@ -441,16 +520,9 @@ def _on_task_tab_changed(self, idx):
         if isinstance(form, MyTasksForm):
             # retrieve the selected task from the form and emit a work-area changed signal:
             task, breadcrumb_trail = form.get_selection()
-            #self._emit_work_area_changed(task, breadcrumb_trail)
```
jfboismenu: This is called by on_selected_task_changed now.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -441,16 +520,9 @@ def _on_task_tab_changed(self, idx):
...
             self._on_selected_task_changed(task, breadcrumb_trail)
-            
+
         elif isinstance(form, EntityTreeForm):
             # retrieve the selection from the form and emit a work-area changed signal:
             selection, breadcrumb_trail = form.get_selection()
-            #selected_entity = selection.get("entity") if selection else None
```
jfboismenu: This is called by on_selected_task_changed now.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -378,22 +405,74 @@ def _on_selected_entity_changed(self, selection_details, breadcrumb_trail):
...
+        :param flat_selection: Array of elements inside the selection.
+        :param workfiles: If True, the method will consider that a sandbox is used
+            if workfiles have user sandboxes.
+        :param publishes: If True, the method will consider that a sandbox is used
+            if publishes have user sandboxes.
+
+        :returns: True if there are user sandboxes for the selection, False otherwise.
+        """
+        # Turn on or off user sandbox filter button.
+        app = sgtk.platform.current_bundle()
```
josh-t: Looks like several places you're grabbing the current bundle, would it make sense to keep a reference to the app on the instance?

robblau: Honestly, it is good practice to not keep your own pointer.  Right now that call always returns that same thing, but what if at a future date it doesn't?  Kind of like keeping a pointer to the shotgun API isn't great since we generate them per thread and it leaves the door open to ending up using the same object across threads.

jfboismenu:
The design has changed quite a bit. Now the button is asynchronously updated. The file finder now emits an event when a work area settings have been loaded but before the sandbox users have been resolved. The code now waits for that event to come it to enable the button if required.

A descriptive tooltip was also added to the button.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -445,33 +472,12 @@ def _on_selected_entity_changed(self, selection_details, breadcrumb_trail):
...
-            # build the work area for this context:
-            work_area = WorkArea(context)
-
-            if work_area.work_area_contains_user_sandboxes and workfiles:
-                return True
-            elif work_area.publish_area_contains_user_sandboxes and publishes:
-                return True
-
-        return False
+    def _on_uses_user_sandboxes(self, work_area):
```
josh-t: missing docs

PR #18, opened 2016-04-22
For #35806, better error reporting

jfboismenu:
- Errors are now more user facing in the file manager app.
- It is now possible to have no entity tabs. (makes sense from the point of view of a user that only uses My Tasks)

UI preview:
screen shot 2016-04-25 at 9 15 53 am


python/tk_multi_workfiles/file_finder.py
```diff
@@ -808,8 +810,12 @@ def _on_background_task_failed(self, task_id, search_id, msg, stack_trace):
             return
         self.stop_search(search_id)
 
+        app = sgtk.platform.current_bundle()
```
jfboismenu: Log the error and callstack to the console instead of sending it all to the UI.

python/tk_multi_workfiles/file_form_base.py
```diff
@@ -282,7 +282,7 @@ def _fileitem_from_path(self, path, work_area):
         :param work_area:   A WorkArea instance representing the work area the file is in
         :returns:           A FileItem representing the specified path in the specified work area
         """
-        if not path or not work_area or not work_area.work_template:
+        if not path or not work_area:
```
jfboismenu: A couple lines under this there's a second test for Falsyness, so I removed this one that was redundant.

python/tk_multi_workfiles/ui/file_group_widget.py
```diff
@@ -106,7 +106,7 @@ def setupUi(self, FileGroupWidget):
 "font: 11px;\n"
 "color: grey;\n"
 "}")
-        self.msg_label.setAlignment(QtCore.Qt.AlignLeading|QtCore.Qt.AlignLeft|QtCore.Qt.AlignVCenter)
+        self.msg_label.setAlignment(QtCore.Qt.AlignLeading|QtCore.Qt.AlignLeft|QtCore.Qt.AlignTop)
```
jfboismenu: Text was centered and it looked really bad when some messages wrapped on multiple lines and some didn't.

python/tk_multi_workfiles/work_area.py
```diff
@@ -1,16 +1,17 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-#
```
jfboismenu: There was a lot of whitespace on non-code lines to I fixed those. Then there were only a few PEP8 errors so I fixed those as well. I will pinpoint where the interesting bits to review are.

python/tk_multi_workfiles/work_area.py
```diff
@@ -190,89 +203,117 @@ def resolve_user_sandboxes(self):
     # Protected methods
 
     def __repr__(self):
-        return ("CTX: %s\n - Work Area: %s\n - Work: %s\n - Publish Area: %s\n - Publish: %s" 
-                % (self._context, 
-                   self.work_area_template, self.work_template, 
+        """
+        Formats a string with information about the work area context's and templates, e.g.::
```
jfboismenu: No code was actually changed, it's only doc.

python/tk_multi_workfiles/work_area.py
```diff
@@ -190,89 +203,117 @@ def resolve_user_sandboxes(self):
...
-            self.save_as_prefer_version_up = resolved_settings.get("saveas_prefer_version_up", False)
-            self.version_compare_ignore_fields = resolved_settings.get("version_compare_ignore_fields", [])
-            extensions = resolved_settings.get("file_extensions") or []
-            extensions = [ext if ext.startswith(".") else ".%s" % ext for ext in extensions if ext]
-            self.valid_file_extensions = extensions
-
-            # test for user sandboxes:
-            self._work_template_contains_user = self.work_template and bool(self._get_template_user_keys(self.work_template))
-            self._publish_template_contains_user = self.publish_template and bool(self._get_template_user_keys(self.publish_template))
+        if self._context:
```
jfboismenu: Of all this block, only the if self._context block is interesting. The rest is jus an indentation change. I removed the try catch so exceptions were not swallowed by the except clause and instead reported back to the caller.

python/tk_multi_workfiles/work_area.py
```diff
@@ -190,89 +203,117 @@ def resolve_user_sandboxes(self):
...
+        self.version_compare_ignore_fields = resolved_settings.get("version_compare_ignore_fields", [])
+        extensions = resolved_settings.get("file_extensions") or []
+        extensions = [ext if ext.startswith(".") else ".%s" % ext for ext in extensions if ext]
+        self.valid_file_extensions = extensions
+
+        # test for user sandboxes:
+        self._work_template_contains_user = self.work_template and bool(self._get_template_user_keys(self.work_template))
+        self._publish_template_contains_user = self.publish_template and bool(self._get_template_user_keys(self.publish_template))
+
+    def are_templates_configured(self):
```
jfboismenu: This is new.

python/tk_multi_workfiles/work_area.py
```diff
@@ -60,12 +61,16 @@ def add(self, context, settings):
...
-    _settings_cache = _SettingsCache() 
-    
+
+    _settings_cache = _SettingsCache()
+
     def __init__(self, ctx=None):
         """
         Construction
+
+        :param ctx: Toolkit context to load the work area settings for.
```
jfboismenu: I also added doc in a lot of places.

python/tk_multi_workfiles/work_area.py
```diff
@@ -190,89 +203,117 @@ def resolve_user_sandboxes(self):
...
             for key in settings_to_find:
                 resolved_settings[key] = app.get_setting(key)
-                
-        else:        
+
+        else:
             # need to look for settings in a different context/environment
             settings = self._get_raw_app_settings_for_context(app, context)
             if not settings:
-                raise TankError("Failed to find Work Files settings for context '%s' in engine instance '%s'.\n\n"
```
jfboismenu: I made the error less technical and used the app's official name.

info.yml
```diff
@@ -100,7 +100,7 @@ configuration:
                       specifies the shotgun entity type to display. *filters* is a list of standard API Shotgun 
                       filters. *hierarchy* is a list of shotgun fields relative to the entity type, defining the 
                       grouping of the tree."
-        allows_empty: False
+        allows_empty: True
```
jfboismenu: There's no reason why a TD shouldn't be able to configure the app to only have the My Tasks tab.

python/tk_multi_workfiles/file_finder.py
```diff
@@ -808,8 +810,12 @@ def _on_background_task_failed(self, task_id, search_id, msg, stack_trace):
...
+        app.log_error(msg)
+        app.log_error(stack_trace)
```
robblau: Should we only log the stack_trace if debug logging is on?

python/tk_multi_workfiles/file_save_form.py
```diff
@@ -134,18 +134,22 @@ def _do_init(self):
...
         current_file = self._get_current_file()
         self._ui.browser.select_work_area(app.context)
         self._ui.browser.select_file(current_file, app.context)
 
-        # initialize the browser with the current file and environment:
-        self._on_browser_file_selected(current_file, env)
+        try:
+            env = WorkArea(app.context)
+        except TankError, e:
+            self._disable_save_and_warn(str(e))
```
robblau: Would the stack trace here be interesting somewhere?

python/tk_multi_workfiles/file_save_form.py
```diff
@@ -301,9 +319,10 @@ def _generate_path(self, env, name, version, use_next_version, ext, require_path
 
         # first make  sure the environment is complete:
         if not env or not env.context:
-            raise TankError("Please select a work area to save into...")
-        elif not env.work_template:
-            raise TankError("The working directory has not been defined for this work area.  Please select another work area.")
+            raise TankError("Please select a work area to save into.")
+        if not env.are_templates_configured():
```
robblau: Possible to have a more specific error that mentions which templates explicitly?
Or change are_tempaltes_configured into something like ```assert_templates_are_configured```?

python/tk_multi_workfiles/file_save_form.py
```diff
@@ -445,9 +464,14 @@ def _on_browser_work_area_changed(self, entity, breadcrumbs):
...
             app = sgtk.platform.current_bundle()
             context = app.sgtk.context_from_entity_dictionary(entity)
-            env = WorkArea(context)
-        self._on_work_area_changed(env)
-        self._start_preview_update()
+
+            try:
+                env = WorkArea(context)
+            except TankError, e:
+                self._disable_save_and_warn(str(e))
```
robblau: Is there an interesting place to put the whole stack trace?  Especially if debug logging is on?

python/tk_multi_workfiles/work_area.py
```diff
@@ -190,89 +203,117 @@ def resolve_user_sandboxes(self):
...
-                                "Please ensure that the Work Files app is installed for the environment that will"
-                                " be used for this context." % (context, app.engine.instance_name))
-            
+                raise TankError(
+                    "Failed to find the Shotgun File Manager settings for this work area.\n\n"
+                    "Please ensure that the app is installed for the environment that will "
+                    "be used for this work area."
```
robblau: Is the environment available?  Would be nice to be able to tell somebody that there were no settings for tk-multi-workfiles2 in tk-maya in shot_step.yml.  Especially when they copy/paste the error in a support ticket.

python/tk_multi_workfiles/file_finder.py
jfboismenu: Good point.

python/tk_multi_workfiles/file_save_form.py
jfboismenu: Actually, the try except is not necessary. I've added a comment why in the code.

python/tk_multi_workfiles/file_save_form.py
jfboismenu: I thought about that. I was trying to strike a balance between having a super technical message and having something that wasn't too technical. I opted for only mentioning the word "template" since by that point a TD will have to go in the yaml, know that the templates are not properly configured and then will look at the 4 values.

jfboismenu: So after reviewing your other comments, I've decided to switch back to more detailed error messages.

python/tk_multi_workfiles/work_area.py
jfboismenu: Unfortunately the environment file name is not available from the public API.

PR #21, opened 2016-05-13
For #36351, gentler error reporting

jfboismenu:
While the last update to error reporting was a lot better than what we had before, we went from no error messages to way to way too alarmous error messages. This branch tries to find some middle ground by being informative and not have people think is something is actually broken and need to be taken care of. 

The approach taken was to first make the messages more neutral. Second, most of these "errors" are actually entirely valid states for the application, in certain contexts. Therefore, the way these warnings are flowing through the code are different, so that they don't get reported as errors by the backend and end up showing in error logs, especially in Maya where the message area turned red even for innocuous messages.

Therefore, all "errors" related to missing templates or missing app in the environment are treated as warnings and take a new code path. Internal errors and YAML parsing errors however are still regarded as actual error and will be reported like before, as they should be.

screen shot 2016-05-13 at 1 35 32 pm
screen shot 2016-05-13 at 12 29 54 pm

### DIFF
--- a/python/tk_multi_workfiles/errors.py
+++ b/python/tk_multi_workfiles/errors.py
@@ -12,6 +12,7 @@
 Workfiles 2 related errors.
 """
 
+from .work_area import WorkArea
 from sgtk import TankError
 
 
@@ -39,7 +40,7 @@ class MissingTemplatesError(WorkfilesError):
         """
         Generates a warning for when templates are not all configured.
         """
-        if len(missing_templates) == 4:
+        if len(missing_templates) == WorkArea.NB_TEMPLATE_SETTINGS:
             return "No templates have been defined."
         else:
             # Then take every template except the last one and join them with commas.
--- a/python/tk_multi_workfiles/errors.py
+++ b/python/tk_multi_workfiles/errors.py
@@ -1,0 +1,103 @@
+# Copyright (c) 2016 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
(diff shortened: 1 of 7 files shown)


python/tk_multi_workfiles/errors.py
```diff
@@ -0,0 +1,63 @@
...
+        :param missing_templates: List of templates that are missing.
+        """
+        WorkfilesError.__init__(self, self.generate_missing_templates_message(missing_templates))
+
+    @classmethod
+    def generate_missing_templates_message(self, missing_templates):
+        """
+        Generates a warning for when templates are not all configured.
+        """
+        if len(missing_templates) == 4:
```
robblau: We shouldn't have 4 here as a magic number... put a constant somewhere for the # of templates or something else that'll be easier to catch if we ever add (or remove) a template from the settings?

python/tk_multi_workfiles/file_list/file_group_widget.py
```diff
@@ -93,7 +94,13 @@ def set_item(self, model_idx):
         if search_status == FileModel.SEARCHING and not idx_has_children:
             search_msg = "Searching for files..."
         elif search_status == FileModel.SEARCH_COMPLETED and not idx_has_children:
-            search_msg = "No files found!"
+            templates = work_area.get_missing_templates()
+            if not work_area.are_settings_loaded():
+                search_msg = "The Shotgun File Manager hasn't been setup."
```
robblau: Last time we talked we decided to standardize on "Workfiles" when we talk about the app rather than a mixture of workfiles and file manager.  Should update to "Shotgun Workfiles hasn't been setup."

jfboismenu: Hum.. I was sure we went the other way around. Should I update info.yml as well then? Could this have any unforseen side-effect in the App Store or Zendesk?

robblau: Just change it here for now, we can update ZD docs in the future... I think there would need to be more than just the info.yml updated.

PR #23, opened 2016-06-01
For #36753, workfiles unexpected switching

jfboismenu:
The main problem this branch fixes is the unexpected context switching when saving a scene. This happened every single time a model was being refreshed when in the file save dialog. To do this, I went through several iterations of code changes, each corresponding to a single commit:

- I've started by refactoring the MyTasksForm and EntityTreeForm, which shared very similar code. MyTasksForm is now a special case of EntityTreeForm and therefore deriving from it.
- I then did some PEP8 checks (but not all) on the edited files because... why not? :)
- Then I fixed the issue where the selection was being updated at the model was being torn down, which forced the entity form selection to keep updating and in the end pick the very first item in the list and effectively changed the context in which we'd save.
- Finally, I fixed a small slowdown introduced by listening to all rows being inserted. We only really need to know when the model has been refreshed to know when the selection can be updated and therefore we're only listening for that event now.

I recommend reviewing this one commit at a time.

### DIFF
--- a/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
+++ b/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
@@ -104,12 +104,12 @@ class EntityTreeForm(QtGui.QWidget):
         entity_model.modelReset.connect(self._model_reset)
 
         if entity_model:
+            entity_model.data_refreshed.connect(self._on_data_refreshed)
 
             if True:
                 # create a filter proxy model between the source model and the task tree view:
                 filter_model = EntityTreeProxyModel(self, ["content", {"entity": "name"}] + extra_filter_fields)
                 monitor_qobject_lifetime(filter_model, "%s entity filter model" % search_label)
-                filter_model.rowsInserted.connect(self._on_filter_model_rows_inserted)
                 filter_model.setSourceModel(entity_model)
                 self._ui.entity_tree.setModel(filter_model)
 
@@ -511,7 +511,7 @@ class EntityTreeForm(QtGui.QWidget):
         # emit selection_changed signal:
         self.entity_selected.emit(selection_details, breadcrumbs)
 
-    def _on_filter_model_rows_inserted(self, parent_idx, first, last):
+    def _on_data_refreshed(self, modifications_made):
         """
         Slot triggered when new rows are inserted into the filter model.  When this happens
         we just make sure that any new root rows are expanded.
@@ -520,6 +520,9 @@ class EntityTreeForm(QtGui.QWidget):
         :param first:       The first row id inserted
         :param last:        The last row id inserted
         """
+        if not modifications_made:
+            return
+
         # expand any new root rows:
         self._expand_root_rows()
 
--- a/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
+++ b/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
@@ -51,7 +51,7 @@ class EntityTreeForm(QtGui.QWidget):
     # Signal emitted when the 'New Task' button is clicked.
     create_new_task = QtCore.Signal(object, object)# entity, step
 
-    def __init__(self, entity_model, search_label, allow_task_creation, extra_filter_fields, parent):
+    def __init__(self, entity_model, search_label, allow_task_creation, extra_fields, parent):
         """
         Construction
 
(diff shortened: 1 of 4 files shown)


python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -44,30 +51,32 @@ def __init__(self, label, entity):
...
 
-    def __init__(self, entity_model, search_label, allow_task_creation, parent):
+    def __init__(self, entity_model, search_label, allow_task_creation, extra_filter_fields, parent):
         """
         Construction
 
-        :param entity_model:    The Shotgun Model this widget should connect to
-        :param search_label:    The hint label to be displayed on the search control
-        :param parent:          The parent QWidget for this control
+        :param entity_model:        The Shotgun Model this widget should connect to
```
thebeeland: Minor: type docs for the params

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -90,17 +99,21 @@ def __init__(self, entity_model, search_label, allow_task_creation, parent):
...
         self._ui.entity_tree.collapsed.connect(self._on_item_collapsed)
 
+        self._is_resetting_model = False
+        entity_model.modelAboutToBeReset.connect(self._model_about_to_reset)
+        entity_model.modelReset.connect(self._model_reset)
+
         if entity_model:
+            entity_model.data_refreshed.connect(self._on_data_refreshed)
 
             if True:
```
thebeeland: Why's this here?

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -113,6 +126,20 @@ def __init__(self, entity_model, search_label, allow_task_creation, parent):
         if selection_model:
             selection_model.selectionChanged.connect(self._on_selection_changed)
 
+    def _model_about_to_reset(self):
+        # Catch the currently selected item and convert it to dictionary form
+        # so we can pick it back after the model is reset.
+        if self._current_item_ref:
```
thebeeland: I'm having trouble finding where `_current_item_ref` is being set to something other than None. If I'm just blind please ignore this!

thebeeland: Fail. I found it now. 👍

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -661,9 +701,3 @@ def _build_breadcrumb_trail(self, idx):
 
         # return reversed list:
         return breadcrumbs[::-1]
```
thebeeland: I did not know a list could be reversed this way. I learned something new!

python/tk_multi_workfiles/my_tasks/my_tasks_form.py
```diff
@@ -1,395 +1,63 @@
...
-
-                self._ui.search_ctrl.search_edited.connect(self._on_search_changed)
-            else:
-                self._ui.task_tree.setModel(tasks_model)
-
-        # connect to the selection model for the tree view:
-        selection_model = self._ui.task_tree.selectionModel()
-        if selection_model:
-            selection_model.selectionChanged.connect(self._on_selection_changed)
+        if tasks_model and True:
```
thebeeland: and True?

python/tk_multi_workfiles/my_tasks/my_tasks_form.py
```diff
@@ -1,395 +1,63 @@
...
-                self._ui.task_tree.selectionModel().clear()
-
-            # detach the filter model from the view:
-            view_model = self._ui.task_tree.model()
-            if view_model:
-                self._ui.task_tree.setModel(None)
-                if isinstance(view_model, EntityProxyModel):
-                    view_model.setSourceModel(None)
-
+            EntityTreeForm.shut_down(self)
```
thebeeland: Minor: I prefer using `super()`. I feel the same way about doing so with `__init__` but it's not the end of the world.

jfboismenu: I'm torn about it too. But in workfiles 2 case, it was originally written not using super so I've followed the pattern.

PR #26, opened 2016-06-29
#37315: Switches to using change_context rather than an engine teardown/restart.

thebeeland:
Using change_context is faster, and appears to resolve some issues we have in 3ds Max.


jfboismenu:
Ship it!

PR #27, opened 2016-09-16
Accounts for receiving datetime.datetime objects from SG.

thebeeland:
This cropped up during Max 2017 compatibility testing. Passing a datetime object to the string formatter results in a TypeError being raised. Handling it here in a backwards-compatible way.


manneohrstrom:
do we know why this is happening? could the reason you are seeing this be because you have invalid caches, created by Josh's version of the shotgun utils fw that contains datetime objects instead of unix times? Is this issue reproducible on an empty cache? Sorry, just wanted to double check before we add the fix - seems odd that this would be exclusive to 3dsmax! If it was an issue, i am surprised we haven't seen it before!

thebeeland:
You're correct. After the cache was cleared this was no longer an issue.

PR #29, opened 2016-09-22
Support advanced filter syntax and make '{context.entity}' work with it

quornian:
Rather than only supporting a list of lists for the filters config item,
support the alternative dict syntax, and recurse so that '{context.entity}'
and similar tokens are replaced at all levels.


quornian:
@jfboismenu is there anything I need to do or can this change be merged as is? Thanks.

jfboismenu:
@quornian Let me get back to you next week. I need to think a bit more about this. There might be something more general we can put either in the model or in core that would benefit every app.

PR #30, opened 2016-09-30
For #38434 and #38433, better context change handling

jfboismenu:
If an open or save hook fails, we now revert back to the original context. Also, when opening a scene, the hook can now return False to indicate that the hook refused to open the file. The context is then reverted as well.

We only use change_context in non-exception safe ways in the open and save action. The New action is correct.

### DIFF
--- a/hooks/scene_operation_tk-shell.py
+++ b/hooks/scene_operation_tk-shell.py
@@ -70,5 +70,5 @@ class SceneOperation(HookClass):
             return True
         elif operation == "open":
             return QtGui.QMessageBox.question(
-                None, "", "Are you sure?", QtGui.QMessageBox.Yes | QtGui.QMessageBox.No
+                None, "", "Are you sure you want to open?", QtGui.QMessageBox.Yes | QtGui.QMessageBox.No
             ) == QtGui.QMessageBox.Yes
--- a/hooks/scene_operation_tk-shell.py
+++ b/hooks/scene_operation_tk-shell.py
@@ -11,19 +11,19 @@
 import os
 
 import sgtk
+from sgtk.platform.qt import QtGui
 
 HookClass = sgtk.get_hook_baseclass()
 
 
 class SceneOperation(HookClass):
     """
-    Hook called to perform an operation with the
-    current scene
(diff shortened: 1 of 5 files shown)


hooks/scene_operation_tk-shell.py
```diff
@@ -68,4 +68,7 @@ def execute(self, operation, file_path, context, parent_action, file_version, re
             return os.getcwd()
         elif operation == "reset":
             return True
-
+        elif operation == "open":
```
jfboismenu: Slightly easier QAing for the app.

hooks/scene_operation_tk-shell.py
```diff
@@ -68,4 +68,7 @@ def execute(self, operation, file_path, context, parent_action, file_version, re
...
+            return QtGui.QMessageBox.question(
+                None, "", "Are you sure?", QtGui.QMessageBox.Yes | QtGui.QMessageBox.No
```
thebeeland: Are we sure of what? If this message makes sense in context, then it's fine. Just making sure...

python/tk_multi_workfiles/actions/file_action.py
```diff
@@ -121,7 +123,29 @@ def change_context(ctx):
...
+    def restore_context(parent_ui, ctx):
+        """
+        Utility method to restore the original context when a file operation failed.
+
+        A dialog will display the error if the restoration fails. This method is exception safe.
+
+        :param PySide.QtWidget parent_ui: Parent for the error dialog, if needed.
+        :param sgtk.Context ctx: Context to restore.
+        """
+        app = sgtk.platform.current_bundle()
```
thebeeland: Why can't this be a non-static method and then just call `self.parent`?

thebeeland: I see now how it's being used in open_file_action. This makes sense.

hooks/scene_operation_tk-shell.py
jfboismenu: This happens when the user clicks save in the hook for tk-shell. This is there only to make the app hooks testable out of the box with the shell engine.

PR #31, opened 2016-10-06
For #38554, sandbox constant refresh fix

jfboismenu:
This fixes an issue where different contexts with different sandboxes would update the sandbox user list in random fashion and the last update won. It took a while to figure this out since user sandboxes are usually created on all steps so the sandboxes are the same for every context. However, Task folders are created on demand and therefore the list of sandboxes per context is different and that's the scenario that caused an issue.

The fix is simple. We'll simply assume that any updates to the sandbox users list is additive and we'll clear the list on selection change.

I recommend skipping the first commit since it is only whitespace changes on comment lines and easy PEP8 fixes.

### DIFF
--- a/python/tk_multi_workfiles/file_filters.py
+++ b/python/tk_multi_workfiles/file_filters.py
@@ -1,33 +1,34 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
 Collection of filters to be used when filtering files in the file views.
 """
 
-import sgtk
 from sgtk.platform.qt import QtCore
 
 from .user_cache import g_user_cache
+
 
 class FileFilters(QtCore.QObject):
     """
     Implementation of the FileFilters class
     """
+
     # signal emitted whenever something in the filters is changed
     changed = QtCore.Signal()
     # signal emitted whenever the available users are changed
     available_users_changed = QtCore.Signal(object)# list of users
     # signal emitted whenever the users changed:
     users_changed = QtCore.Signal(object)# list of users
-    
+
     def __init__(self, parent):
         """
         Construction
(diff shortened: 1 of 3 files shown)


python/tk_multi_workfiles/file_model.py
```diff
@@ -983,35 +981,11 @@ def _on_finder_work_area_resolved(self, search_id, work_area):
...
-        entity_key = self._gen_entity_key(search.entity)
-
-        # keep track of work area for this entity:
-        self._entity_work_area_map[entity_key] = work_area
-
-        # consolidate list of sandbox users for all current work areas:
-        have_user_sandboxes = False
-        users_by_id = {}
-        handled_entities = set()
-        for search in self._in_progress_searches.values():
```
jfboismenu: This is the main piece of code that was wrong before. It accumulated the sandboxes for all contexts that were being scanned. I can't understand why it was written this way, but I definitely say the the logic was wrong because the intent is to accumulate the list of all users across sandboxes and this accumulates the list of all users across sandboxes that are being scanned. As jobs complete, this list shortens and less and less sandboxes will be returned. The new code simply broadcasts the users that were found.

python/tk_multi_workfiles/file_filters.py
```diff
@@ -1,86 +1,109 @@
...
-    #@property
+    def _reset_user_lists(self):
+        self._available_users = [g_user_cache.current_user] if g_user_cache.current_user else []
+        self._users = [g_user_cache.current_user] if g_user_cache.current_user else []
+
+    def add_users(self, users):
+        """
+        Adds to the list of available user sandboxes.
+        """
+        nb_users_before = len(self._available_users)
```
josh-t: super minor: `num_` instead of `nb_`?

python/tk_multi_workfiles/file_filters.py
```diff
@@ -1,86 +1,109 @@
...
+
+        # merge the two lists, discarding doubles.
+        new_users_by_id = {user["id"]: user for user in users}
+        available_users_by_id = {user["id"]: user for user in self._available_users}
+        available_users_by_id.update(new_users_by_id)
+        self._available_users = available_users_by_id.values()
+
+        # Something was added to the set.
+        if nb_users_before != len(self._available_users):
```
josh-t: I don't think I know enough about how workfiles works to understand this scenario completely. Is this always an additive thing? i.e., could the number be the same and the users be different?

python/tk_multi_workfiles/file_model.py
```diff
@@ -98,7 +98,7 @@ def data(self, role=QtCore.Qt.UserRole+1):
                 # just return the default implementation:
                 return QtGui.QStandardItem.data(self, role)
 
-        def setData(self, value, role=QtCore.Qt.UserRole+1):
+        def setData(self, value, role=QtCore.Qt.UserRole + 1):
```
josh-t: I realize this is a minor style fix, but I think we should move the role + 1 to a class variable instead of doing it here in the data methods.

python/tk_multi_workfiles/file_model.py
```diff
@@ -158,7 +157,7 @@ def data(self, role=QtCore.Qt.UserRole+1):
                 # just return the default implementation:
                 return FileModel._BaseModelItem.data(self, role)
 
-        def setData(self, value, role=QtCore.Qt.UserRole+1):
+        def setData(self, value, role=QtCore.Qt.UserRole + 1):
```
josh-t: same comment about role + 1.

python/tk_multi_workfiles/file_model.py
```diff
@@ -272,7 +271,7 @@ def data(self, role=QtCore.Qt.UserRole+1):
...
```
josh-t: role + 1 again.

python/tk_multi_workfiles/file_filters.py
jfboismenu: Yes, it is additive. I'll clarify the comment.

PR #35, opened 2016-12-14
Changes to workfiles to work with new defer loaded sg model

manneohrstrom:
Previously, workfiles assumed that the tree structure on the left hand side was always fully loaded. The new optimized shotgun model defer loads stuff into memory for performance and this caused workfiles to not get all the child data correctly when an item in the left hand side tree was clicked. 
This fix adds a recursive load of child data just in time before it's needed, ensuring consistent behaviour when running with the new optimized sg model.

The filtering logic has been updated to ensure the entire tree is fetched just in time prior to filtering.

This change is backwards compatible and will not affect users running against older versions of the sg model but it is required if you want to run with the latest optimized version.

### DIFF
--- a/python/tk_multi_workfiles/entity_proxy_model.py
+++ b/python/tk_multi_workfiles/entity_proxy_model.py
@@ -25,27 +25,12 @@ class EntityProxyModel(HierarchicalFilteringProxyModel):
         HierarchicalFilteringProxyModel.__init__(self, parent)
         self._compare_fields = compare_sg_fields
 
-    def __ensure_subtree_loaded(self, model_index):
-        """
-        Recursively descend and ensure all data is
-        fetched for the given model index.
-
-        :param model_index: Model index to process
-        """
-        model = model_index.model()
-        if model.canFetchMore(model_index):
-            model.fetchMore(model_index)
-        for child_index in range(model.rowCount(model_index)):
-            child_model_index = model.index(child_index, 0, parent=model_index)
-            self.__ensure_subtree_loaded(child_model_index)
-
     def setFilterFixedString(self, pattern):
         """
         Overriden base class method to set the filter fixed string
         """
         # ensure model is fully loaded before we attempt any searching
-        root_index = self.sourceModel().invisibleRootItem().index()
-        self.__ensure_subtree_loaded(root_index)
+        self.sourceModel().ensure_data_is_loaded()
 
         # call base class
         return super(EntityProxyModel, self).setFilterFixedString(pattern)
@@ -55,8 +40,7 @@ class EntityProxyModel(HierarchicalFilteringProxyModel):
         Overriden base class method to set the filter regular expression
         """
         # ensure model is fully loaded before we attempt any searching
-        root_index = self.sourceModel().invisibleRootItem().index()
-        self.__ensure_subtree_loaded(root_index)
+        self.sourceModel().ensure_data_is_loaded()
 
         # call base class
         return super(EntityProxyModel, self).setFilterRegExp(reg_exp)
--- a/python/tk_multi_workfiles/entity_proxy_model.py
+++ b/python/tk_multi_workfiles/entity_proxy_model.py
@@ -25,6 +25,42 @@ class EntityProxyModel(HierarchicalFilteringProxyModel):
         HierarchicalFilteringProxyModel.__init__(self, parent)
         self._compare_fields = compare_sg_fields
 
+    def __ensure_subtree_loaded(self, model_index):
+        """
+        Recursively descend and ensure all data is
+        fetched for the given model index.
+
+        :param model_index: Model index to process
+        """
+        model = model_index.model()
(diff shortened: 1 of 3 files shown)


jfboismenu:
I'm worried that this is going to become a recurring pattern for clients of the model, who will now have to assume something is not loaded before digging into the hierarchy. Maybe we can think of some sort of accessor API?

Also, by looking at change, I'm worried that the framework changes actually means we will be breaking clients applications. Maybe this should be a major revision for the framework?

python/tk_multi_workfiles/entity_proxy_model.py
```diff
@@ -25,6 +25,42 @@ def __init__(self, parent, compare_sg_fields=None):
...
+        """
+        Recursively descend and ensure all data is
+        fetched for the given model index.
+
+        :param model_index: Model index to process
+        """
+        model = model_index.model()
+        if model.canFetchMore(model_index):
+            model.fetchMore(model_index)
+        for child_index in range(model.rowCount(model_index)):
```
josh-t: I'm probably forgetting something here, but if `fetchMore` is asynchronous, won't `rowCount` be unreliable?

python/tk_multi_workfiles/entity_proxy_model.py
```diff
@@ -25,6 +25,42 @@ def __init__(self, parent, compare_sg_fields=None):
...
+    def setFilterFixedString(self, pattern):
+        """
+        Overriden base class method to set the filter fixed string
+        """
+        # ensure model is fully loaded before we attempt any searching
+        root_index = self.sourceModel().invisibleRootItem().index()
+        self.__ensure_subtree_loaded(root_index)
+
+        # call base class
+        return super(EntityProxyModel, self).setFilterFixedString(pattern)
```
josh-t: Will the filter be applied dynamically as items are added to the model?

python/tk_multi_workfiles/entity_proxy_model.py
```diff
@@ -25,6 +25,42 @@ def __init__(self, parent, compare_sg_fields=None):
...
+    def setFilterRegExp(self, reg_exp):
+        """
+        Overriden base class method to set the filter regular expression
+        """
+        # ensure model is fully loaded before we attempt any searching
+        root_index = self.sourceModel().invisibleRootItem().index()
+        self.__ensure_subtree_loaded(root_index)
+
+        # call base class
+        return super(EntityProxyModel, self).setFilterRegExp(reg_exp)
```
josh-t: Same question here. Will this update dynamically as items are added?

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -315,6 +315,17 @@ def _get_entity_details(self, idx):
         if not idx.isValid():
             return {}
 
+        # first, ensure that all child data has been loaded
+        def __fetch_more_r(model_index):
```
josh-t: Any reason this method is in the class? Perhaps it could just be a top-level function in the file?

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -315,6 +315,17 @@ def _get_entity_details(self, idx):
...
+            """Recursively descend and ensure all data is fetched"""
+            model = model_index.model()
+            if model.canFetchMore(model_index):
+                model.fetchMore(model_index)
+            for child_index in range(model.rowCount(model_index)):
```
josh-t: Same question here about `fetchMore` being asynchronous.

python/tk_multi_workfiles/entity_proxy_model.py
manneohrstrom: yep, that's all taken care of by the higher level proxy model logic

python/tk_multi_workfiles/entity_proxy_model.py
manneohrstrom: yep, that's all taken care of by the higher level proxy model logic

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
manneohrstrom: it's a super custom method meant for this purpose only. But i can move it out if you feel that improves readability!

manneohrstrom:
@jfboismenu all good and valid points. We are fundamentally changing how the model behaves, although its interfaces remain the same. We have seen that our own apps make use of the fact that the entire tree is loaded in to memory in various ways, so it's possible that client code will break if similar assumptions have been made there. Maybe we can discuss? 

Perhaps one way to do this would be to bump the major version to 5 and add an `expand_tree` method to `ShotgunModel` that could be called from these various places? It just further increases the cost of this change. Let's chat about it today!

manneohrstrom:
Bumped to shotgunmodel v5 and addressed comments. Much cleaner now!

PR #42, opened 2017-05-05
#42449 Usability tweaks to improve working with large amounts of tasks

pscadding:
#42449
#42445
#39615

Made some changes that hopefully make working with large amounts of tasks a bit easier.
More work is still needed, but these can be considered quick fixes.

Added option to disable auto expansion of root level in the tree views.

- You can now set a`auto_expand_tree` key on the app config to true or false, to toggle this behaviour

Added option to set a list of status that the My Tasks tab will use as a filter to exclude.

- can now set a config key for `my_tasks_filter_statuses: ['fin','omt']` to filter out tasks from the "my tasks" tab, with the status codes listed.

Changed tree view search box behaviour so that it only performs the search after hitting enter.
This helps when performing the search is very slow, as it will now only perform the search after enter is hit, rather than every character entered.

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -140,15 +140,14 @@ configuration:
       default_value: []
 
     my_tasks_filters:
-      type: dict
+      type: list
       description: List of filters to apply to the tasks in the My Tasks tab.
       values:
-        type: list
+        type: shotgun_filter
       allows_empty: True
       default_value:
-        filters:
-          - [task_assignees, is, '{context.user}']
-          - [sg_status_list, not_in, [fin,omt]]
+        - [task_assignees, is, '{context.user}']
+        - [sg_status_list, not_in, [fin,omt]]
 
     file_extensions:
         type: list
--- a/info.yml
+++ b/info.yml
@@ -129,7 +129,7 @@ configuration:
     auto_expand_tree:
         type: bool
         description: Automatically expands the root level of the treeviews.
-        default_value: True
+        default_value: False
 
     my_tasks_extra_display_fields:
       type: list
--- a/info.yml
+++ b/info.yml
@@ -126,6 +126,11 @@ configuration:
         default_value: True
         default_value_tk-hiero: False
 
+    auto_expand_tree:
+        type: bool
+        description: Automatically expands the root level of the treeviews.
+        default_value: True
+
     my_tasks_extra_display_fields:
       type: list
       description: List of additional fields to display with the task in the My Tasks list.
@@ -133,6 +138,14 @@ configuration:
         type: str
       allows_empty: True
       default_value: []
+
+    my_tasks_filter_statuses:
+      type: list
+      description: List of additional fields to display with the task in the My Tasks list.
+      values:
+        type: str
+      allows_empty: True
+      default_value: [fin,omt]
 
     file_extensions:
         type: list
--- a/info.yml
+++ b/info.yml
@@ -139,13 +139,16 @@ configuration:
       allows_empty: True
       default_value: []
 
-    my_tasks_filter_statuses:
-      type: list
-      description: List of additional fields to display with the task in the My Tasks list.
+    my_tasks_filters:
+      type: dict
+      description: List of filters to apply to the tasks in the My Tasks tab.
       values:
-        type: str
+        type: list
       allows_empty: True
-      default_value: [fin,omt]
+      default_value:
+        filters:
+          - [task_assignees, is, '{context.user}']
+          - [sg_status_list, not_in, [fin,omt]]
 
     file_extensions:
         type: list
--- a/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
+++ b/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
@@ -80,13 +80,14 @@ class EntityTreeForm(QtGui.QWidget):
 
         # load the setting that states whether the first level of the tree should be auto expanded
         app = sgtk.platform.current_bundle()
-        self._auto_expand_tree = app.get_setting("auto_expand_tree", True)
+        self._auto_expand_tree = app.get_setting("auto_expand_tree")
 
         # set up the UI
         self._ui = Ui_EntityTreeForm()
         self._ui.setupUi(self)
 
         self._ui.search_ctrl.set_placeholder_text("Search %s" % search_label)
+        self._ui.search_ctrl.setToolTip("Press enter to complete the search")
 
         # enable/hide the my-tasks-only button if we are showing tasks:
         have_tasks = (entity_model and entity_model.get_entity_type() == "Task")
--- a/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
+++ b/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
@@ -78,6 +78,10 @@ class EntityTreeForm(QtGui.QWidget):
         self._expanded_items = set()
         self._auto_expanded_root_items = set()
 
+        # load the setting that states whether the first level of the tree should be auto expanded
+        app = sgtk.platform.current_bundle()
+        self._auto_expand_tree = app.get_setting("auto_expand_tree", True)
+
         # set up the UI
         self._ui = Ui_EntityTreeForm()
         self._ui.setupUi(self)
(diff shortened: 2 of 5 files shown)


python/tk_multi_workfiles/file_form_base.py
```diff
@@ -121,11 +121,13 @@ def _build_my_tasks_model(self):
 
         # get any extra display fields we'll need to retrieve:
         extra_display_fields = app.get_setting("my_tasks_extra_display_fields")
+        my_tasks_filter_statuses = app.get_setting("my_tasks_filter_statuses",["fin"])
```
jfboismenu: Same comment here about the default value. Especially since the default you are providing here is different than the one from `info.yml`

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -117,7 +121,7 @@ def __init__(self, entity_model, search_label, allow_task_creation, extra_fields
                 self._ui.entity_tree.setModel(filter_model)
 
                 # connect up the filter controls:
-                self._ui.search_ctrl.search_edited.connect(self._on_search_changed)
+                self._ui.search_ctrl.search_changed.connect(self._on_search_changed)
```
jfboismenu: I wonder if we need some sort of visual indication in the UI that the search will only start when you hit enter now.

I think currently the placeholder says "Search..."

Maybe we should update it to something "Type something and hit enter to search..." or something.

info.yml
```diff
@@ -134,6 +139,14 @@ configuration:
       allows_empty: True
       default_value: []
 
+    my_tasks_filter_statuses:
+      type: list
+      description: List of additional fields to display with the task in the My Tasks list.
```
jfboismenu: Bad description

python/tk_multi_workfiles/my_tasks/my_tasks_model.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, project, user, extra_display_fields, parent, bg_task_manager=
         """
         self.extra_display_fields = extra_display_fields or []
         filters = [["project", "is", project],
-                   ["task_assignees", "is", user]]
+                   ["task_assignees", "is", user],
```
jfboismenu: Instead of allowing people to simply take over the status filter, I think we should simply have the whole filter be an option in the yml file, just like we do for the entity tabs. This would allow people to do whatever they want.

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
jfboismenu: Btw, I think my suggestion is quite bad, so don't use that! :p

python/tk_multi_workfiles/my_tasks/my_tasks_model.py
pscadding: @jfboismenu, great point, might as well give more control. Do you think we should use the same syntax as the filters for the other tabs? https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/master/info.yml#L109

jfboismenu:
Yes, my thoughts exactly, including the support for using the {context.user} notation. You should be able to refactor the functionality quite easily.




On Fri, May 5, 2017 at 3:51 PM -0400, "Philip Scadding" <notifications@github.com<mailto:notifications@github.com>> wrote:


@pscadding commented on this pull request.

________________________________

In python/tk_multi_workfiles/my_tasks/my_tasks_model.py<https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/42#discussion_r115075159>:

> @@ -39,7 +39,8 @@ def __init__(self, project, user, extra_display_fields, parent, bg_task_manager=
         """
         self.extra_display_fields = extra_display_fields or []
         filters = [["project", "is", project],
-                   ["task_assignees", "is", user]]
+                   ["task_assignees", "is", user],


@jfboismenu<https://github.com/jfboismenu>, great point, might as well give more control. Do you think we should use the same syntax as the filters for the other tabs? https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/master/info.yml#L109

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
manneohrstrom: maybe a tooltip?

python/tk_multi_workfiles/my_tasks/my_tasks_model.py
JeanChristopheMorin: +1 for full control on this. The more control users have, and the happier they will be ;)

PS: Sorry to intervene in your PR, I don't know if we are allowed or not to do so...

jfboismenu: Hey @JeanChristopheMorin , good to see you! Don't hesitate to chime in on our PRs if you think something is good or bad. We want to make you all happy!

jfboismenu: @pscadding Yes, my thoughts exactly, including the support for using the {context.user} notation. You should be able to refactor the functionality quite easily.

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -78,6 +78,10 @@ def __init__(self, entity_model, search_label, allow_task_creation, extra_fields
         self._expanded_items = set()
         self._auto_expanded_root_items = set()
 
+        # load the setting that states whether the first level of the tree should be auto expanded
+        app = sgtk.platform.current_bundle()
+        self._auto_expand_tree = app.get_setting("auto_expand_tree", True)
```
pscadding: Yes! removing the default arg of True is fine, it is pulling the value from the info yml if not supplied in the env config.

python/tk_multi_workfiles/entity_tree/entity_tree_form.py
pscadding: I've added a tooltip, which isn't a bad idea, but I don't think everyone will think to check that when they realise it doesn't search automatically. I can't really think of a great solution.

We could have a setting in the config, but that really does feel a bit obscure and over the top.

pscadding:
I have pushed some new changes.

I have added the ability to completely specify the filters used on the My Tasks tab. the Project filter is still hard coded to be inserted. I don't however like the way I have had to get the info.yml settings to work for it. I would have prefered to have something like

```
my_tasks_filters:
      type: list
      description: List of filters to apply to the tasks in the My Tasks tab.
      values:
        type: list
      allows_empty: True
      default_value: 
        - [task_assignees, is, '{context.user}']
        - [sg_status_list, not_in, [fin,omt]]
```

but have had to settle for this:

```
my_tasks_filters:
      type: dict
      description: List of filters to apply to the tasks in the My Tasks tab.
      values:
        type: list
      allows_empty: True
      default_value:
        filters:
          - [task_assignees, is, '{context.user}']
          - [sg_status_list, not_in, [fin,omt]]
```

Haven't addressed the autoexpand comment on whether it should be false by default, waiting for comment from @robblau 

Phil

pscadding:
**Update**

`my_task_filters` now uses `shotgun_filter` as type, which solves the problem with the ugly nested dict. Now looks like this. (thanks to @manneohrstrom )

```
my_tasks_filters:
        type: list
        values:
            type: shotgun_filter
        allows_empty: True
        default_value:
        - [task_assignees, is, '{context.user}']
        - [sg_status_list, not_in, [fin,omt]]
```

info.yml
```diff
@@ -126,6 +126,11 @@ configuration:
         default_value: True
         default_value_tk-hiero: False
 
+    auto_expand_tree:
+        type: bool
+        description: Automatically expands the root level of the treeviews.
+        default_value: True
```
robblau: Hey all, sorry missed this until Phil's recent update.

I agree we should set the default value to False.  We weren't even going to offer this as an option in the ticket I remember for this.  :)

pscadding: OK I have set it False now. Rob says it was never intended to be an option, I can remove the option if we feel that is best?

python/tk_multi_workfiles/util.py
```diff
@@ -261,3 +261,38 @@ def get_template_user_keys(template):
         if key.shotgun_entity_type == "HumanUser":
             user_keys.add(key.name)
     return user_keys
+
+
+def resolve_filters(filters):
+    """
+    
+    When passed a list of filters, it will resolve strings found in the filters using the context
+    example: '{context.user}' could get resolved to {'type': 'HumanUser', 'id': 86, 'name': 'Philip Scadding'}
```
jfboismenu: We should take the time at some point to refactor this in shotgunutils @manneohrstrom . Shall I create a ticket?

manneohrstrom: yes pls @jfboismenu !

jfboismenu: Done!

PR #44, opened 2017-08-25
Step is not a valid entity supported by path_cache, so attempting to …

mbrainerd:
…find files owned by Step will always fail.


jfboismenu:
Hi @mbrainerd ,
Could you explain the circumstances under which the issue you are trying to fix arises?

mbrainerd:
Sure no problem. In our default setup, the user launches the DCC from the shot directory and Toolkit initializes its context from the path. The "Pipeline Step" is also set via an environment variable configured on a per user basis, so the Toolkit context will have its "Project", "Entity", "HumanUser", and "Step" attributes populated on launch, but not "Task".

When the workfiles "Load" toolkit app is opened, the FileBrowser modal immediately attempts a file search based on the highest ranking entity defined in the current context (i.e. the logic in browser_form.py:219). The current logic starts with "Task", which in our case isn't defined, and then falls back to "Step" and then "Entity".

Based on the logic starting at tank.folder.operations.py:149, "Step" is not supported in the folder creation process (i.e. path_cache population), which is later used by the modal to handle file lookups, so will always be empty. Additionally, and more critically, the logic in the subsequent functions that utilize "ctx_entity" defined in "select_work_area()" attempts to create a new context based on that single entity. Since "Step" entities do not have a direct link to the Sequence/Shot/Asset entities in which they are being utilized, this always results in a malformed context, which causes an error to occur during our call to get_environment later on in the code, among other issues. Fortunately, this is not a critical bug as the first thing a user does is select a "Task" from the left hand menu, which then works properly as "Task" will result in a more well-formed context, it just presents an error message to the user when first running the app. Also, IMHO, it feels as though this logic should be reworked to reuse the context already provided by the host App rather than attempting to derive a new context for a single entity in that context, but that would clearly be a much larger refactor. 

Please let me know if I can help explain this more clearly.

Best,
Mike

PR #46, opened 2017-10-20
removed infinite while loop in folder creation

MattRickS:
Workfiles FileAction would hang indefinitely as the template it was trying to resolve was always pointing back to the same one.


pscadding:
Hi Matt

Verry sorry we've not seen this or addressed it. It so happens that we also discovered this issue recently and have fixed it. We're about to release a fix, probably today.
Apologies again for not getting to your PR, we'll try and do better next time. Also if you feel like its not been noticed, feel free to drop a ticket into our support, and we'll try and get eyes on it.

Regards
Phil

jfboismenu:
Released as part of v0.11.5.

PR #51, opened 2018-01-09
38483 Tech review for step filtering and deferred queries.

sgsteph:
**Shouldn't be merge on master.** the goal of this PR is to get a tech review.

Two new features were added:
1) Step filtering for Tasks, allowing to reduce the number of data which is pulled from Shotgun by only retrieving Tasks for a subset of Pipeline Steps.
2) Deferred queries, allowing to have a primary query, and then dynamically pull children when an item is expanded in the tree view.

A `ShotgunUpdatableEntityModel` class deriving from `ShotgunEntityModel` was added, and it contains most of the logic dealing with these two new features.
When step filtering is changed, the `ShotgunEntityModel` SG filters are updated and the full data set is reloaded, if there is no deferred query attached to the model. If there is a deferred query, the data is re-fetched with the update filters for all expanded items in the tree view. For this latter case, there is certainly something which must be done if there is a lot of expanded items as we're doing this synchronously.

Some logic was added to maintain the selection and the tree expansion when a model is reset. Before, it only worked for leaves, it now works for any item in the tree. Please note that in certain cases I was not able to leverage existing members or methods from the Shotgun utils base classes because they were private. This is something that should be addressed, either by adding accessors or moving some code on the base implementations.

As I went through the code, I added some comments, docstrings, reformatted the code to make it easier to understand. I moved the styling I found into an external style sheet file as well.

So the goal of this PR is to assess if there is any red flag which should be raised with the proposed approach from a tech point of view, and if it would make sense to move part of this code into `tk-framework-shotgunutils` rather than having some custom classes in workfile2.


sgsteph:
An example of the settings with deferred queries and the classic Tasks tree:
```
settings.tk-multi-workfiles2.launch_at_startup:
  allow_task_creation: true
  create_new_task_hook: "{self}/create_new_task.py"
  custom_actions_hook: "{self}/custom_actions.py"
  entities:

  - caption: Assets (deferred)
    entity_type: Asset
    hierarchy: [sg_asset_type, code]
    sub_hierarchy:
        entity_type: Task
        filters:
        link_field: entity
        fields: [step]

  - caption: Assets Tasks
    entity_type: Task
    filters:
    - [entity, type_is, Asset]
 #   - [sg_status_list, is, ip]
    hierarchy: [entity.Asset.sg_asset_type, entity, step, content]

  - caption: Shots (deferred)
    entity_type: Shot
    filters:
    - [code, is_not, null]
    hierarchy: [sg_sequence, code]
    sub_hierarchy:
        entity_type: Task
        filters:
        link_field: entity
        fields: [step]

  - caption: Shots Tasks
    entity_type: Task
    filters:
    - [entity, type_is, Shot]
#    - [sg_status_list, is, ip]
    hierarchy: [entity.Shot.sg_sequence, entity, step, content]
  file_extensions: []
  hook_copy_file: "{self}/copy_file.py"
  hook_filter_publishes: default
  hook_filter_work_files: default
  hook_scene_operation: default
  launch_at_startup: true
  my_tasks_extra_display_fields: []
  saveas_default_name: scene
  saveas_prefer_version_up: false
  show_my_tasks: true
  template_publish:
  template_publish_area:
  template_work:
  template_work_area:
  version_compare_ignore_fields: []
  location: "@apps.tk-multi-workfiles2.location"

```

python/tk_multi_workfiles/entity_model.py
```diff
@@ -0,0 +1,485 @@
+# Copyright (c) 2015 Shotgun Software Inc.
```
sgsteph: This is where `ShotgunUpdatableEntityModel` is defined.

manneohrstrom:
Adding @jfboismenu as he is the one in the team who has had the highest level of exposure to this code so far!

manneohrstrom:
starting now - sorry about the delay.

sgsteph:
Answering a couple of questions from @jfboismenu:
* I feel like the `update_filters` method could be added to the shotgun util base implementation. So the `ShotgunUpdatableEntityModel` would only contain the code specific to deferred queries (and would be renamed to something else).
* I thought about allowing multiple levels of deferred queries, but if we want to go that route, I think that this should be added to shotgun utils. Settings would be slightly different: a list of SG queries to go from one level to the other, with a _pivot_ field allowing to link one level to the previous one. And certainly some kind of _cards_ view instead of displaying a tree which is not well suited for this kind of approach.

Let's have a catchup with Manne and you and talk about it!

sgsteph:
Closing this PR in favor of #52.

PR #53, opened 2018-02-26
38483 step filter qa fix

sgsteph:
Addressed problems reported in QA:

* Fixed step _select_ buttons styling by using an arbitrary color, rather than having the background transparent, which causes styling issues in various DCCs.
* Revisited tree expansion saving/restoring without using `weakref` which could cause some C++ object deleted errors, and ensured the expansion is correctly restored when the model is reset.
* Fixed some refresh problems with the current selection and ensured the file lists is always accurate with what is shown in the tree: if a step/task is filtered out, its files will not appear in the right pane anymore.
* Restored app defaults to _classic_ mode with the `step_filter_on` setting to narrow down the list of step filters shown in entity tabs.

Please note that a fix is needed from tk-framework-shotgunutils: https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/84.

The required  version will be updated once the framework is out.


style.qss
```diff
@@ -29,7 +29,7 @@ QWidget {
 #step_filters_frame QPushButton {
     font-size: 9px;
     margin: 5px;
-    background-color: transparent;
+    background-color: #515151;
```
manneohrstrom: nitpick: is there a way we could avoid adding color constants to the QSS? Can we rely on the underlying style to get it? What is the normal bg col for `QPushButton`? Maybe worth adding a comment explaining why we need to set this to a dark gray.

info.yml
```diff
@@ -114,23 +114,17 @@ configuration:
           type: dict
         default_value:
         - caption: Assets
-          entity_type: Asset
+          entity_type: Task
```
manneohrstrom: please add a comment here to point out that for backwards comp reasons, the default value isn't using the most optimal setup etc (and maybe what that would look like). Some kind of breadcrumb or little heads up just to remind users i think would be useful.

python/tk_multi_workfiles/entity_models/deferred_model.py
```diff
@@ -85,6 +85,7 @@ def __init__(self, entity_type, filters, hierarchy, fields, deferred_query, *arg
         self._deferred_query = deferred_query
         self._task_step_icons = {}
         self._deferred_models = {}
+        self._pending_delayed_data_refreshed = False
```
manneohrstrom: i suggest adding a comment to briefly explain what state we are tracking. (It's not clear to be what *pending delayed data* is.)

python/tk_multi_workfiles/entity_models/deferred_model.py
```diff
@@ -519,6 +520,23 @@ def _on_deferred_data_refreshed(self, sg_entity, changed, pending_refresh=False)
                 # The item might be already gone because one its ancestor was
                 # deleted.
                 self._delete_item(item)
+        # If we have any uid which is not in both sets, then things changed.
+        if existing_uids ^ refreshed_uids:
```
manneohrstrom: For clarity, i suggest using the verbose form here: [`s.symmetric_difference(t)`](https://docs.python.org/2/library/sets.html). I suspect most readers of the code are not familiar with what the `^` operator is doing for a set.

python/tk_multi_workfiles/entity_models/deferred_model.py
```diff
@@ -519,6 +520,23 @@ def _on_deferred_data_refreshed(self, sg_entity, changed, pending_refresh=False)
                 # The item might be already gone because one its ancestor was
                 # deleted.
                 self._delete_item(item)
+        # If we have any uid which is not in both sets, then things changed.
```
manneohrstrom: what if it's in one set but not the other? Does that not mean that things have changed?

python/tk_multi_workfiles/entity_models/deferred_model.py
```diff
@@ -519,6 +520,23 @@ def _on_deferred_data_refreshed(self, sg_entity, changed, pending_refresh=False)
...
+        if existing_uids ^ refreshed_uids:
+            # Make sure we only trigger a refresh only once after all pending
+            # deferred data has been processed by posting a single data_refreshed
+            # emission in the event queue after all events which are already queued.
+            if not self._pending_delayed_data_refreshed:
```
manneohrstrom: this feels a bit entangled. Do you think there would be a way to encapsulate it into an object for clarity? That object would hold the `_pending_delayed_data_refreshed` state internally and all we need to do was to emit the signal and implement the slot and the logic for making sure that it only fired once was kept within the object. Just something to consider to reduce the complexity of the runtime state of this class.

style.qss
sgsteph: I changed it to `none` and it seems to behave (Maya, Nuke, Photoshop).

python/tk_multi_workfiles/entity_models/deferred_model.py
sgsteph: will clarify the comment: the difference returns the list of uids which are either in one or the other, but not both. Given that `symmetric_difference` is as obscured as `^` to me, I will keep the operator, but have a better comment.

python/tk_multi_workfiles/entity_models/deferred_model.py
sgsteph: I quickly tried the suggested object approach, but this makes things slightly more complicated, and in the end we would still need a member to store the object (if I understood well). So I added a method to encapsulate this code. We now simply have `self._post_delayed_data_refreshed()`, which, hopefully, makes the code clear enough and not so entangled.

PR #54, opened 2018-03-05
38483 step filter 3ds  and step filter fixes

sgsteph:
* Fixed a refresh problem in 3ds 2016 by adding an overlay widget to tree views and showing/hiding it which forces a refresh of the UI. Otherwise, the current tree view was only redrawn when moving the mouse and the focus changed.
* Fixed broken step _select_ buttons styling in 3ds 2018 by removing the `margin` styling and adding extra space in the horizontal layout containing them. This is not as nice as previously, but it is pretty hard to get something which works in all DCCs.
* Fixed a bug in the step filtering: previously the current selection was retrieved from the widgets. However, because these widgets are created on the fly when needed, depending on the order of user actions we could end up with only half of the actual selection, because the user didn't check the tab containing these steps, and so the widgets were not created.


style.qss
```diff
@@ -29,8 +29,7 @@ QWidget {
 
 #step_filters_frame QPushButton {
     font-size: 9px;
-    margin: 5px;
-    background-color: none;
+    background-color: #none;
```
sgsteph: This is the fix for the styling in 3ds 2018.

python/tk_multi_workfiles/step_list_filter.py
```diff
@@ -124,6 +124,11 @@ def select_all_steps(self, value=True):
             # Only toggle widgets currently active
             if not self._list_widget.isRowHidden(item_row):
                 item = self._list_widget.item(item_row)
+                item_step = item.data(QtCore.Qt.UserRole)
```
manneohrstrom: normally we always have a constant here to scope what we are actually storing e..g.

```python
item.data(QtCore.Qt.UserRole+self.PIPELINE_STEP_ID)
```

Either that, or a comment outlining what's being stored would be good.

python/tk_multi_workfiles/step_list_filter.py
```diff
@@ -224,21 +223,26 @@ def _ensure_widgets_for_entity_type(self, entity_type):
...
-            widget.toggled.connect(self._on_step_filter_toggled)
+            widget.toggled.connect(lambda value, step_id=step["id"] : self._on_step_filter_toggled(step_id, value))
             item = QtGui.QListWidgetItem("", self._list_widget)
             item.setData(QtCore.Qt.UserRole, step)
             self._list_widget.setItemWidget(item, widget)
             self._step_widgets[entity_type].append(widget)
         return self._step_widgets[entity_type]
 
-    def _on_step_filter_toggled(self, value=None):
+    def _on_step_filter_toggled(self, step_id, value=None):
```
manneohrstrom: `value` seems like an odd name for a boolean parameter - could we rename it to something clearer?

PR #55, opened 2018-03-07
For #38483, improved refresh.

sgsteph:
Couple of fixes related to refreshing the model:

* In classic mode, emit a data changed as soon as some cached data is loaded, instead of waiting for the async refresh to trigger a data changed callback.
* Fixed the overlay widget by using the right widget and the right parent.
* Refresh the file browser when no entities are found after a deferred model refresh.

### DIFF
--- a/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
+++ b/python/tk_multi_workfiles/entity_tree/entity_tree_form.py
@@ -86,10 +86,6 @@ class EntityTreeForm(QtGui.QWidget):
         self._expanded_item_values = []
         self._selected_item_value = []
 
-        # An overlay widget we can hide and show when the model is being refreshed.
-        # This is mainly used as a workaround for 3ds 2016 refresh problems: by
-        # hiding the widget when the data is refreshed, the UI is properly refreshed.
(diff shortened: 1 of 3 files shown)


python/tk_multi_workfiles/entity_tree/entity_tree_form.py
```diff
@@ -627,9 +627,8 @@ def _on_data_refreshed(self, modifications_made):
         # try to select the current entity from the new items in the model:
         prev_selected_item = self._reset_selection()
         self._update_selection(prev_selected_item, True)
-        # Hide the overlay widget, if it was visible
-        if not self._refresh_overlay_widget.isHidden():
-            self._refresh_overlay_widget.hide()
+        # Hide the overlay widget
+        self._refresh_overlay_widget.hide()
```
sgsteph: Removed the `isHidden` test as it didn't seem to work, so we now systematically hide the overlay widget, even if it's already hidden. Any worry, let me know!

josh-t: seems fine. i like explict. :)

PR #57, opened 2018-04-16
Feature/tab settings

mrharris:
This is an implementation of https://support.shotgunsoftware.com/hc/en-us/community/posts/360002740574-workfiles2-settings-for-main-panel-tabs

There were too many entries in the "All" tab for it to be a reasonable default. Most of the entries were publishes that did not need to be opened very often. Artists ended up getting confused by which file they should be opening

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -206,6 +206,14 @@ configuration:
         description: Controls whether new tasks can be created from the app.
         default_value: True
 
+    file_browser_tabs:
+        type: list
+        description: "Controls what tabs are visible in the main browser"
+        allows_empty: True
+        values:
+          type: str
+        default_value: ['All', 'Working', 'Publishes']
+
     # Save specific options
     #
 
--- a/python/tk_multi_workfiles/browser_form.py
+++ b/python/tk_multi_workfiles/browser_form.py
@@ -225,7 +225,7 @@ class BrowserForm(QtGui.QWidget):
             if "All" in preferred_tabs:
                 self._add_file_list_form("All", "All Files", show_work_files=True, show_publishes=True)
             if "Working" in preferred_tabs:
-                self._add_file_list_form("Workingxxx", "Work Files", show_work_files=True, show_publishes=False)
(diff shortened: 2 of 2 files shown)


info.yml
```diff
@@ -206,6 +206,14 @@ configuration:
         description: Controls whether new tasks can be created from the app.
         default_value: True
 
+    file_browser_tabs:
+        type: list
+        description: "Controls what tabs are visible in the main browser"
+        allows_empty: True
```
jfboismenu: Please set to `False`, otherwise it allows someone to configure the app without any tabs, which doesn't work.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -221,9 +221,13 @@ def set_models(self, my_tasks_model, entity_models, file_model):
             self._file_model.set_users(self._file_filters.users)
 
             # add an 'all files' tab:
-            self._add_file_list_form("All", "All Files", show_work_files=True, show_publishes=True )
-            self._add_file_list_form("Working", "Work Files", show_work_files=True, show_publishes=False)
-            self._add_file_list_form("Publishes", "Publishes", show_work_files=False, show_publishes=True)
+            preferred_tabs = app.get_setting("file_browser_tabs")
+            if "All" in preferred_tabs:
```
jfboismenu: Instead of looking for values that work, it would be great it we could loop over all of them and compare them using `lower` on the user's input to make it a bit more permissive. For entries that do not map to any valid values, we should log a warning using a logger's `warning` method.  You can check out our doc on how to [get a Toolkit logger](https://developer.shotgunsoftware.com/tk-core/utils.html#logging-from-within-your-app-engine-or-framework) and then use it.

josh-t:
I've addressed JF's comments in a branch on our side. We'll get this reviewed, merged, and QA'd very soon. Closing this PR. Here's the new one: https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/62.

PR #59, opened 2018-05-29
SG-4520: Fixes the "my tasks" checkbox in entity tree views.

thebeeland:
In order for the "My Tasks" checkbox to function properly, the entity tree data has to have been queried from Shotgun. Otherwise, there are no Task entities to check assignments on and we end up with an empty view even when there are Tasks assigned to the current user. 

**Symptoms:**

- Navigating to a tab with an entity tree view ("Shots" or "Assets" in the default configuration) and checking on the "My Tasks" checkbox resulted in an empty entity tree.
- If the user drilled down into the entity tree prior to checking on "My Tasks", thereby forcing the tree to be populated, the "My Tasks" checkbox would work properly for any Tasks that had already been drilled down to.

**Solution:**

When filtering for "My Tasks" behavior, we force child item data to be loaded for each index processed, which ensures we have all of the data from Shotgun about Tasks that we need to determine their assignments.

**Caveats:**

I assume this could be slow if there's a ton of entity data to load. However, I don't see any other way for it to work, so we probably have to live with it.

### DIFF
--- a/python/tk_multi_workfiles/entity_tree/entity_tree_proxy_model.py
+++ b/python/tk_multi_workfiles/entity_tree/entity_tree_proxy_model.py
@@ -57,6 +57,13 @@ class EntityTreeProxyModel(EntityProxyModel):
             if not src_idx.isValid():
                 return False
 
+            # The only way we can traverse down to the Task level of the
+            # entity hierarchy and determine whether a Task is assigned to
+            # the current user is to load the child data up front for each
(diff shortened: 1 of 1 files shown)


thebeeland:
It does, but I suspect it's either this or we remove the feature entirely.

thebeeland:
We're not going this route, as it's not an efficient solution.

PR #62, opened 2018-08-06
For #SG-7629: Fixes for PR from BoulderMedia

josh-t:
Addresses PR comments from JF here:

https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/57/files

* allows_empty is now False
* loops over valid tabs and compares against lower() for
  configured values.
* warns/raises now if tab names are problematic

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -206,6 +206,17 @@ configuration:
         description: Controls whether new tasks can be created from the app.
         default_value: True
 
+    file_browser_tabs:
+        type: list
+        description: "A list of tab names that are visible in the main browser. Values
+                      can be any combination of the following tab names: 'All',
+                      'Working', and 'Publishes'. The "
+
+        allows_empty: False
+        values:
+          type: str
+        default_value: ['All', 'Working', 'Publishes']
+
     # Save specific options
     #
 
--- a/info.yml
+++ b/info.yml
@@ -210,8 +210,8 @@ configuration:
         type: list
         description: "A list of tab names that are visible in the main browser. Values
                       can be any combination of the following tab names: 'All',
-                      'Working', and 'Publishes'. The "
-
+                      'Working', and 'Publishes'. The order of the specified
+                      values determines the order of the tabs in the app."
         allows_empty: False
         values:
           type: str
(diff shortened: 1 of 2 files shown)


info.yml
```diff
@@ -206,6 +206,17 @@ configuration:
...
+    file_browser_tabs:
+        type: list
+        description: "A list of tab names that are visible in the main browser. Values
+                      can be any combination of the following tab names: 'All',
+                      'Working', and 'Publishes'. The "
+
+        allows_empty: False
+        values:
+          type: str
+        default_value: ['All', 'Working', 'Publishes']
```
thebeeland: I'm assuming the order in the list determines the order the tabs are shown in the UI? If so, the first value specifies what the user sees by default when they open the app? If that's the case, can that be explained better in the description above?

python/tk_multi_workfiles/browser_form.py
```diff
@@ -220,10 +245,50 @@ def set_models(self, my_tasks_model, entity_models, file_model):
...
+            # lowercase tabs to display as configured for the app
+            tabs_to_display = [
+                t.lower() for t in app.get_setting("file_browser_tabs")]
+
+            tab_count = 0
+
+            # iterate over each configured tab and see if it is valid (in the
+            # list of valid tabs)
+            for tab in tabs_to_display:
+                if tab in self.TAB_INFO.keys():
```
thebeeland: I don't think you need the keys() call here. You can just do `tab in self.TAB_INFO`

info.yml
```diff
@@ -206,6 +206,17 @@ configuration:
         description: Controls whether new tasks can be created from the app.
         default_value: True
 
+    file_browser_tabs:
+        type: list
+        description: "A list of tab names that are visible in the main browser. Values
+                      can be any combination of the following tab names: 'All',
+                      'Working', and 'Publishes'. The "
```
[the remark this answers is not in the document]
josh-t: i think i was about to explain the ordering and got sidetracked and forgot.

info.yml
josh-t: yep. see previous comment.

PR #67, opened 2019-01-16
Replaced toTuple for PyQt compatibility

j0yu:
Discovered while trying to integrate `tk-katana`.

See [PyQt API docs](http://pyqt.sourceforge.net/Docs/PyQt4/qcolor.html) compared to [PySide API docs](https://srinikom.github.io/pyside-docs/PySide/QtGui/QColor.html#PySide.QtGui.PySide.QtGui.QColor.toTuple) for QColor

### DIFF
--- a/app.py
+++ b/app.py
@@ -149,9 +149,10 @@ class MultiWorkFiles(sgtk.platform.Application):
         """
         Color used to display errors in the UI.
 
-        :returns: An RGBA tuple.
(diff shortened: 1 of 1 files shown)


jfboismenu:
Thanks, @j0yu ! We've logged this and will have a look!

thebeeland:
Thanks, @j0yu! I've pulled this into #70 and we'll get it through QA and released from there.

PR #68, opened 2019-01-16
Added __ignore *args

j0yu:
Discovered while trying to integrate `tk-katana`.

Due to signals/slots, Qt was passing in extra args to `show_file_open_dlg()` and `show_file_save_dlg()` when the menu actions are triggered, resulting in Python complaining about extra args given when only `self` is expected.


j0yu:
Will handle inside `tk-katana` instead, closing

PR #69, opened 2019-01-24
For SG-10271, unique version number across sandboxes

jfboismenu:
Takes Amy's fix for sandboxes and generalizes it so any time a workfile is saved a unique version number is generated across sandboxes.

This required a new field to be introduced on the `Workfile` custom entity: `sg_name_field`. This is the value of the name field in the save dialog. I didn't go with simply `sg_name` because I was afraid it would cause confusion between `code`, `name` and `sg_name`.

When moving to this new version of the app, the Workfile entity first need to be migrated to populate that field.

It could be (I haven't actually executed this) as simple as

```
for entity in sg.find("CustomEntity45", []):
    # Create a fake publish payload so we can ask core to resolve its path like we do
    # in the workfiles management app.
    fake_sg_publish_data = {
            "id": entity["id"],
            "path": entity["sg_path"]
    }
    sgtk.util.resolve_publish_path(sgtk, fake_sg_publish_data);
    sg.update(
        entity['type'], entity['id']
        {'sg_name_field': template.get_fields()["name"]}
   )
```
which will go through every single workfile, extract the name from the path field and push back that name to shotgun.

### DIFF
--- a/hooks/workfiles_management.py
+++ b/hooks/workfiles_management.py
@@ -85,7 +85,10 @@ class WorkfilesManagement(sgtk.get_hook_baseclass()):
             "sg_path": sg_path,
             "project": context.project
         }
-
+        if "name" in work_template.keys:
+            new_workfile["sg_name_field"] = work_template.get_fields(path)["name"]
+        else:
+            new_workfile["sg_name_field"] = None
         if self._is_using_sandboxes(work_template):
             new_workfile["sg_sandbox"] = context.user
 
@@ -97,6 +100,26 @@ class WorkfilesManagement(sgtk.get_hook_baseclass()):
             self.WORKFILE_ENTITY,
             new_workfile
         )
+
+    def get_next_workfile_version(self, name, context, work_template):
+
+        # TODO: This could probably be optimized by a summarize call for the maximum
+        # value for the field version.
+        results = self._find_workfile_entities(
+            context,
+            work_template,
+            ["sg_version"],
+            # We want unique version number per user.
+            filter_by_user=False,
+            name_filter=name
+        )
+
+        # max fails on empty iterables.
+        if not results:
+            return 1
+
+        # Find the maximum version number
+        return max(results, key=lambda entity: entity["sg_version"])["sg_version"] + 1
 
     def find_work_files(self, context, work_template, version_compare_ignore_fields, valid_file_extensions):
         """
@@ -214,7 +237,7 @@ class WorkfilesManagement(sgtk.get_hook_baseclass()):
                 return True
         return False
 
-    def _find_workfile_entities(self, context, work_template, fields, filter_by_user):
+    def _find_workfile_entities(self, context, work_template, fields, filter_by_user, name_filter=None):
         """
         Find all workfiles entities for a given template and context.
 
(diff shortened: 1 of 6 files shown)


python/tk_multi_workfiles/actions/interactive_open_action.py
```diff
@@ -214,28 +214,36 @@ def _open_workfile(self, file, env, parent_ui):
                     ctx_fields = local_ctx.as_template_fields(env.work_template)
                     fields.update(ctx_fields)
                     if "version" in fields:
-                        # Check for existing Workfiles linked to the current context for the
```
jfboismenu: This was Amy's code.

python/tk_multi_workfiles/actions/open_file_action.py
```diff
@@ -240,13 +240,18 @@ def _open_in_current_work_area(self, src_path, src_template, file, src_work_area
...
-        # Q. should the next version include the current version?
-        all_versions = [v for v, f in file_versions.iteritems()] + [file.version]
-        max_version = max(all_versions)
-        self._version = max_version+1
+
+        if next_version is not None:
+            self._version = next_version
+        else:
+            # Q. should the next version include the current version?
+            all_versions = [v for v, f in file_versions.iteritems()] + [file.version]
```
manneohrstrom: can you turn `[v for v, f in file_versions.iteritems()]` into just `file_versions.keys()`?

hooks/workfiles_management.py
```diff
@@ -31,6 +31,7 @@ class WorkfilesManagement(sgtk.get_hook_baseclass()):
     WORKFILE_ENTITY = "CustomEntity45"
     # The custom entity has the following custom fields:
     # sg_version (number): version number of the scene
+    # sg_name_field (text): user specified name at save time
```
manneohrstrom: what is the difference between this and `CustomEntity45.code` - i am not sure i follow! May be worth explaining.

hooks/workfiles_management.py
```diff
@@ -85,7 +86,10 @@ def register_workfile(self, name, version, context, work_template, path, descrip
             "sg_path": sg_path,
             "project": context.project
         }
-
+        if "name" in work_template.keys:
+            new_workfile["sg_name_field"] = work_template.get_fields(path)["name"]
```
manneohrstrom: what is the difference between `work_template.get_fields(path)["name"]` and the input `name` parameter?

hooks/workfiles_management.py
jfboismenu: I'll add a comment at the bottom explaining the difference!

hooks/workfiles_management.py
```diff
@@ -41,12 +42,17 @@ class WorkfilesManagement(sgtk.get_hook_baseclass()):
...
+    # The sg_name_field value however is the value of the {name} token inside the file name,
+    # e.g. given the template {name}.v{version}.ma and the file name (code) of car.v003.ma,
+    # sg_name_field would be set to "car".
 
     def register_workfile(self, name, version, context, work_template, path, description, image):
         """
         Register a work file with Shotgun.
 
-        :param name str: Name of the work file.
+        :param name str: File name on disk of the work file. Not to be confused with the {name} token.
```
manneohrstrom: @jfboismenu why not rename the `name` parameter to be `filename` for clarity? If you are leaving it for compatibility reasons, let's discuss! There aren't yet that many users of this (and the main user submitted this PR hehe). If you chose name for symmetry reasons to match other code, it's a different discussion and then we probably shouldn't mess with renaming it.... :)

PR #73, opened 2019-04-10
Add a find_work_files hook for perforce work flows

jhultgre:
Add a hook to allow finding additional workfiles.
This is handy in a perforce setup where all the workfiles might not have been synced to the users machine when it runs the _find_workfiles method.

### DIFF
--- a/hooks/find_work_files.py
+++ b/hooks/find_work_files.py
@@ -1,0 +1,29 @@
+# Copyright (c) 2015 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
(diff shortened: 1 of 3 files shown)


hooks/find_work_files.py
```diff
@@ -0,0 +1,29 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+
+import sgtk
+
+HookClass = sgtk.get_hook_baseclass()
+
+
+class FilterWorkFiles(HookClass):
```
BoubQC: class should be named FindWorkFiles.

hooks/find_work_files.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Hook that can be used to filter the list of work files found by the app for the current
```
BoubQC: Description should be updated.

PR #74, opened 2019-04-18
SG-9846 Add Hooks to allow publishes and versions to have a badge in the UI

willis102:
Add hooks to allow custom badges to be applied to publishes and versions in the UI.


hooks/get_badge.py
```diff
@@ -0,0 +1,100 @@
...
+class GetBadge(HookClass):
+    """
+    Hook that can be used to generate badges for display on publishes and work files.
+    """
+
+    def get_publish_badge(self, publish_details, publish_path, **kwargs):
+        """
+        Generate a badge for a publish.
+
+        :param publish_details:         Dictionary
```
jfboismenu: the right syntax to employ here is `:param dict publish_details` and `:param str publish_path`

The types will be properly referenced in the Sphinx doc.

hooks/get_badge.py
```diff
@@ -0,0 +1,100 @@
...
+                                badge will be generated using that color.
+        """
+        # the default implementation always returns None.
+        return None
+
+    def get_work_file_badge(self, work_file_details, work_file_path, **kwargs):
+        """
+        Generate a badge for a work file.
+
+        :param work_file_details:       Dictionary
```
jfboismenu: same comment here for types.

hooks/get_badge.py
```diff
@@ -0,0 +1,100 @@
...
+        :param badge_color:     QColor
+                                The color of the badge to generate a pixmap for.
+
+        :returns:               A QPixmap of the badge to be used.
+        """
+        # We want to multiply the color onto the (white) badge_default dot to
+        # generate a nice looking badge.
+        badge = QtGui.QPixmap(":/tk-multi-workfiles2/badge_default.png")
+        painter = QtGui.QPainter()
+        painter.begin(badge)
```
jfboismenu: Could we update this to
```python
painter.begin(badge)
try:
        painter.setCompositionMode(QtGui.QPainter.CompositionMode_SourceIn)
        painter.fillRect(badge.rect(), badge_color)
finally:
        painter.end()
```

I've seen hardcrashes in the past when `painter.end` is not invoked properly.

python/tk_multi_workfiles/file_item.py
```diff
@@ -213,6 +220,73 @@ def _set_versions(self, value):
...
+        """
+        return self._badge
+
+    def _set_badge(self, value):
+        """
+        :param value:   The QPixmap to be used as a badge when displaying this file
+                        to the user.
+        """
+        self._badge = value
+    badge = property(_get_badge, _set_badge)
```
jfboismenu: Using the property method is now an anti-pattern. You can simply use the `@property` and `@badge.setter` decorators. I think this was required back in the Python 2.5 because property setters didn't exist.

As long as your class is a new-style class, which `FileItem` is because it derives from `object`, `@badge.setter` will work.

python/tk_multi_workfiles/file_list/file_widget.py
```diff
@@ -112,6 +128,15 @@ def set_thumbnail(self, thumb):
             thumb = QtGui.QPixmap(":/tk-multi-workfiles2/thumb_empty.png")
         self._ui.thumbnail.setPixmap(thumb)
 
+    def set_badge(self, badge):
+        if isinstance(badge, QtGui.QPixmap):
+            self._badge_icon.setPixmap(badge)
+            self._badge_icon.setVisible(True)
+        else:
+            # If we don't have a QPixmap badge, clear and hide the badge.
```
jfboismenu: Could you elaborate why we wouldn't have a QPixmap image? Should we raise an error or log something?

hooks/get_badge.py
willis102: ✅

hooks/get_badge.py
willis102: ✅

hooks/get_badge.py
willis102: Good catch, thanks.  ✅

python/tk_multi_workfiles/file_item.py
willis102: ✅

python/tk_multi_workfiles/file_list/file_widget.py
willis102: Improved the comment.  This could be None or QColor, if there is no badge or the badge hasn't been generated yet.  In either case, we should not show the badge.  Improved comment should help clarify.

PR #78, opened 2019-09-17
Fix bootstrap error on Houdini deadline jobs.

reformstudios:
Houdini deadline jobs uses hython.exe which has no gui and thus no Qt, which workfiles2 needs to load.
This update detects if the app is being loaded in to a houdini gui environment and skips initialising the app if this isn't the case.
This allows houdini to bootstrap SG on deadline without erroring.

### DIFF
--- a/app.py
+++ b/app.py
@@ -24,6 +24,15 @@ class MultiWorkFiles(sgtk.platform.Application):
         """
         Called as the application is being initialized
         """
+        
+        if self.engine.name == "tk-houdini":
+            # When Houdini is bootstrapped on a renderfarm, hython is used which has no gui causing workfiles2 to fail.
+            # Prevent loading of this app if no gui is detected.
+
+            import hou
+            if not hasattr(hou, 'ui'):
+                return
+        
         self._tk_multi_workfiles = self.import_module("tk_multi_workfiles")
         self.__is_pyside_unstable = None
 


app.py
```diff
@@ -24,6 +24,15 @@ def init_app(self):
         """
         Called as the application is being initialized
         """
+        
+        if self.engine.name == "tk-houdini":
```
pscadding: Rather than doing this, it would be better I think to do something like
```
if not self.engine.has_ui:
   return
```
A bit like it's done [here](https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/a65f8d4da2e2c6d6aa98f663e83295e95d234902/app.py#L92), that makes it engine agnostic. 
Also that line I just linked could then lose the UI check, as you would be breaking out sooner if didn't have a UI.

reformstudios:
Oh nice, thanks for the suggestion.

app.py
thebeeland: Phil's correct here. With that done I feel like this is good to go. Like Phil thought, this seems like the right place to put this logic.

thebeeland:
I've gone ahead and made the change that @pscadding suggested and have PR'd it as #79. We'll take it from here @reformstudios, thanks so much!

reformstudios:
No problemo!

PR #79, opened 2019-10-08
SG-14243: Adds a has_ui check before allowing the app to initialize.

thebeeland:
This is just cleaning up what's in #78 to make it a more generalized solution.

### DIFF
--- a/app.py
+++ b/app.py
@@ -28,7 +28,7 @@ class MultiWorkFiles(sgtk.platform.Application):
         self.__is_pyside_unstable = None
 
         if not self.engine.has_ui:
-            self.log_debug("The engine reports that there is no UI. Workfiles2 will not continue initializing.")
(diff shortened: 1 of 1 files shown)


app.py
```diff
@@ -27,6 +27,10 @@ def init_app(self):
         self._tk_multi_workfiles = self.import_module("tk_multi_workfiles")
         self.__is_pyside_unstable = None
 
+        if not self.engine.has_ui:
+            self.log_debug("The engine reports that there is no UI. Workfiles2 will not continue initializing.")
```
pscadding: Good call on the log line, however I think we should avoid this method call specifically, as it is [deprecated](https://developer.shotgunsoftware.com/tk-core/platform.html?highlight=log_debug#sgtk.platform.Application.log_debug).

Should it be `self.logger.debug(...` instead?

PR #84, opened 2020-01-28
SG-13276: python 3 port

jfboismenu:
Adds Python 3 support to workfiles 2. Due to the heavy processing done by the file model, I've kept the iterators on the dictionaries instead of just changing everything to `.values()/.items()/.keys()` like we do elsewhere.

Once again, there's nothing much to say about this port, except that I've added some unit tests for the app. Since the application is broken down into so many small pieces, it's easy to unit test them one at a time, so I've chose to unit those that deal with user sandboxes in the template and schema, since these are workflows that we have a harder time testing usually. I've introduced a base test for the class that both the user menu and file model class use.


python/tk_multi_workfiles/file_list/user_filter_menu.py
```diff
@@ -120,8 +122,8 @@ def _set_selected_users(self, users):
     def _get_available_users(self):
         available_users = set(
             [
-                details.user
-                for details in self._available_users.values()
+                user_id
```
jfboismenu: This method was not called anywhere in the code base and when I tried to invoke it in my tests it failed because `dict` is not a hashable type, so I fixed the method the same way we implemented selected_users_ids.

python/tk_multi_workfiles/file_model.py
```diff
@@ -41,15 +42,15 @@ class SearchDetails(object):
...
+        def __init__(self, name=None, entity=None, child_entities=None, is_leaf=False):
             """
             :param name:    The string name to be used for the search.  This is used when constructing
                             the group name in the model.
             """
             self.name = name
-            self.entity = None
-            self.child_entities = []
-            self.is_leaf = False  # TODO: this does not seem to be used anywhere?
+            self.entity = entity
```
jfboismenu: Makes it easier to create test data if everything can be specified on init.

resources/task_widget.ui
```diff
@@ -26,10 +26,6 @@
 	border-bottom-color: rgb(64,64,64);
 }
 
-#background[selected=false] {
```
jfboismenu: Not needed and caused warnings with PySide2.

python/tk_multi_workfiles/entity_models/deferred_model.py
```diff
@@ -479,10 +479,7 @@ def _on_deferred_data_refreshed(self, sg_entity, changed, pending_refresh=False)
         for sg_deferred_entity in sg_deferred_entities:
             # Update existing items or create new ones.
             uids = self._add_deferred_item_hierarchy(
-                parent_item,
-                deferred_query["hierarchy"],
-                name_field,
-                sg_deferred_entity,
+                parent_item, deferred_query["hierarchy"], name_field, sg_deferred_entity
```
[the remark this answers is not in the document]
jfboismenu: Yup, it's black. It probably figured that the parameters fit on a single line.

PR #85, opened 2020-02-16
use partial instead of lambda for menu actions

jhultgre:
Our custom engine that uses the unreal qt5.6a framework had issues where the right click menu actions wouldn't trigger. Swapping the lambda with a partial fixed the issue.


julien-lang:
Hi @jhultgre, the issue should now be fixed in [v0.18.0](https://github.com/shotgunsoftware/tk-multi-workfiles2/releases/tag/v0.18.0).

PR #91, opened 2020-06-29
SHOT-3416 Added tk-sketchbook to supported Engines list.

rob-aitchison:


rob-aitchison:
Hey @jfboismenu , sorry about the messy commit. Duly noted and I will run the pre-commits from now on.

workfiles does run on startup as expected in tk-sketchbook but it hangs the software launch unless it is declared here in app.py. Is that what you are asking?

Thanks!

jfboismenu:
Hi Rob I'm afraid I'm not sure I understand what you mean here:
> workfiles does run on startup as expected in tk-sketchbook but it hangs the software launch unless it is declared here in app.py
The `SUPPORTED_ENGINES` constants lists DCCs for which `launch_at_startup` will actually work. Anyone can turn the setting to on in their config, but only a subset of engines support it due to certain DCCs being unstable on startup.

Are you saying that tk-sketchbook becomes woobly if you do NOT show the workfiles dialog?

barbara-darkshot:
hey @rob-aitchison, I have a question :-) Why did you add the tk-sketchbook hook inside the app itself instead of putting it in the engine like all the other recent engines? Is it because it's required by the Shotgun/Sketchbook integration?

rob-aitchison:
Hi @barbara-darkshot , it is actually in both places for the short-term. I added it here in order to use the launch_at_startup since SketchBook does not launch by just declaring rut_at_startup as true for a Project context. Once we solve that issue, I will remove it from here, thanks for checking!

rob-aitchison:
I made modifications in the config to no longer requires this.

PR #92, opened 2020-07-03
SG-17473 Increases the maximum save version value to 9999

pscadding:
It now allows the user to save a file with a version number of up to 9999.
It was requested by users that we allow files with three decimal places, I figured giving four gives a bit more headroom just in case.


pscadding:
I've upped the version so it is higher: `9999999` although it does make the version input box a little wider, its still fine.

PR #95, opened 2020-09-17
Proof of concept of work files 2 being used to change context

jfboismenu:
This is a proof of concept. It doesn't actually work at the moment, as the Open button does not get enabled when you pick a task and even if you clicked on it it wouldn't work as you expect. It simply lays out the experience right.


jfboismenu:
@pscadding is implementing this feature based loosely on this POC. No need to keep this around.

PR #96, opened 2020-09-25
WF2 File Open UI Automation

mathurf:
This is the work to have WF2 UI Automation running on Azure


azure-pipelines.yml
```diff
@@ -14,7 +14,7 @@ resources:
     - repository: templates
       type: github
       name: shotgunsoftware/tk-ci-tools
-      ref: refs/heads/master
+      ref: switch_human_user
```
jfboismenu, line 17: If your fix works you should merge it to master and restore the previous ref.

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,449 @@
...
+        "LocalStorage", local_storage["id"], {"windows_path": local_storage["path"]}
+    )
+
+    # Make sure there is not already an automation project created
+    filters = [["name", "is", "Toolkit File Open UI Automation"]]
+    existed_project = sg.find_one("Project", filters)
+    if existed_project is not None:
+        sg.delete(existed_project["type"], existed_project["id"])
+
+    # Create a new project with the Film VFX Template
```
jfboismenu, line 52: Which template?

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,449 @@
...
+        "project": {"type": new_project["type"], "id": new_project["id"]},
+        "code": "seq_001",
+        "sg_status_list": "ip",
+    }
+    new_sequence = sg.create("Sequence", sequence_data)
+
+    # Create a new shot
+    shot_data = {
+        "project": {"type": new_project["type"], "id": new_project["id"]},
+        "sg_sequence": {"type": new_sequence["type"], "id": new_sequence["id"]},
```
jfboismenu, line 70: you can pass sequence_data and project directly here instead of extracting type and id.

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,449 @@
...
+        "code": "shot_001",
+        "sg_status_list": "ip",
+        "task_template": {"type": "TaskTemplate", "id": 5},
```
jfboismenu, line 73: We should look for that task template by it's name and not just 5.

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,449 @@
...
+    sg.create("Shot", shot_data)
+
+    # Create a new asset
+    asset_data = {
+        "project": {"type": new_project["type"], "id": new_project["id"]},
+        "code": "AssetAutomation",
+        "description": "This asset was created by the File Open UI automation",
+        "sg_status_list": "ip",
+        "sg_asset_type": "Character",
+        "task_template": {"type": "TaskTemplate", "id": 1},
```
jfboismenu, line 84: Same here.

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,449 @@
...
+    app_dialog.root.outlineitems["Character"].mouseDoubleClick()
+    app_dialog.root.outlineitems["AssetAutomation"].waitExist(timeout=30)
+    app_dialog.root.outlineitems["AssetAutomation"].mouseDoubleClick()
+    app_dialog.root.outlineitems["Model"].waitExist(timeout=30)
+
+    # Validate content dialog
+    assert app_dialog.root.cells[
+        "AssetAutomation"
+    ].exists(), "AssetAutomation is missing in content dialog"
+    # These tests are failing on Azure but succeed locally. Need more investigation
```
jfboismenu, line 299: instead of commenting them out, you could do

`if "CI" not in os.environ:`

and put your code inside the if.

jfboismenu: This way it would keep running locally.

jfboismenu (approved):
Looks mostly good to me, I only have some nitpicks. Good job.

tests/test_file_open_gui.py
mathurf: Yeah I know but the thing is, I looked at different Shotgun site and it turns out that depending on the unified template version used when the site gets created, task templates can have different names but they always have the same id.

tests/test_file_open_gui.py
mathurf: oups wrong description.

tests/test_file_open_gui.py
mathurf: same reason than above

tests/test_file_open_gui.py
mathurf: oh nice!

tests/test_file_open_gui.py
mathurf: Now using custom task templates

mathurf: Now using custom task templates

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,348 @@
...
+        self.root = parent["Shotgun: File Open"].get()
+
+    def exists(self):
+        """
+        ``True`` if the widget was found, ``False`` otherwise.
+        """
+        return self.root.exists()
+
+    def close(self):
+        self.root.buttons["Close"].get().mouseClick()
```
pscadding, line 114: Could these be broken out into a separate python file and then imported so that they can be reused. The code looks like it is duplicated in multiple files here.

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,348 @@
...
+
+    # Make sure all buttons are showing up
+    assert app_dialog.root.buttons[
+        "+ New Task"
+    ].exists(), "+ New Task button is missing"
+    assert app_dialog.root.buttons[
+        "+ New File"
+    ].exists(), "+ New File button is missing"
+    assert app_dialog.root.buttons["Cancel"].exists(), "Cancel button is missing"
+    assert app_dialog.root.buttons["Open"].exists(), "Open button is missing"
```
pscadding, line 155: We discussed setting up hooks to assert whether the open operation was called correctly. 
Did you look into that? This is all good stuff anyway I just wondered if it was possible test that the hook was called with the correct operation.

tests/test_file_open_gui.py
```diff
@@ -0,0 +1,348 @@
...
+    ].exists(), "Breadcrumb widget is not set correctly"
+
+    # Enable My Tasks Only and make sure Model task is not showing up anymore
+    app_dialog.root.checkboxes["My Tasks Only"].mouseClick()
+    assert (
+        app_dialog.root.outlineitems["Model"].exists() is False
+    ), "Model task shouldn't be visible"
+
+
+def test_shots_tab(app_dialog):
```
pscadding, line 261: It looks like the shot and asset tabs do the same thing really, just with Asset/Shot and different task names.
Perhaps this could be refactored into a single test_tab method, and parametrize like the Python console does here:
https://github.com/shotgunsoftware/tk-multi-pythonconsole/blob/master/tests/test_standalone_app.py#L137-L150

tests/test_file_save_gui.py
```diff
@@ -0,0 +1,364 @@
...
+        "New Task"
+    ].exists(), "New Task task should be visible"
+
+
+def test_shots_tab(app_dialog):
+    """
+    Shot tab UI validation
+    """
+    # Select the Shots tab
+    app_dialog.root.tabs["Shots"].mouseClick()
```
pscadding, line 282: At a quick glance it looks like some of this at the top is the same as the open tab. I wonder if some of this could be put in a fixture that could be shared?

pscadding (approved):
It looks really good, nice work.
I have a few suggestions, around code de-duplication, let me know if you would like a hand.

tests/conftest.py
```diff
@@ -0,0 +1,183 @@
...
+
+import pytest
+import os
+from tk_toolchain.authentication import get_toolkit_user
+
+
+@pytest.fixture(scope="session")
+def shotgun():
+    """
+    Getting credentials from TK_TOOLCHAIN
```
jfboismenu, line 19:
```suggestion
    Provides a connection to Shotgun.
```

jfboismenu (approved):
I agree with Phil. It looks like some of the code would gain from a small refactor.

tests/test_file_open_gui.py
mathurf: Indeed. Fixed

mathurf: and thanks for your help :)

tests/test_file_open_gui.py
mathurf: Not for now, this will need us to be able to bootstrap the config to be able to pass `sgtk.sgtk_from_path()` or `sgtk.sgtk_from_path()`
Maybe a good idea for a next innovation day :)

tests/test_file_save_gui.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+# Parametrize decorator to run the same functions for Assets and Shots tabs.
+@pytest.mark.parametrize(
+    "tab_name, selection_hierarchy, entities",
+    [
+        (
+            "Assets",
+            ("Character", "AssetAutomation", "Model"),
+            ("Asset", "AssetAutomation", "Model", "Rig"),
```
pscadding, line 76: Perhaps a small comment on these explaining what they are and why the order is important?

I know I provided the selection_hierarchy as an example so I know that the order if the names of the tree items in the selection order, but it's not clear to others.
And the entities also appear to be in some kind of hierarchal order, but If I was to want to introduce another tab check for say Sequences then I'm not sure I'd know what to put here and what it is expecting. E.i. does it always have to be four items, like Entity Type. Entity Name, Step name, task name?

Are both model and rig tasks? If so then maybe it's work breaking those out into a tasks parameter?

tests/workfiles2_functions.py
```diff
@@ -0,0 +1,196 @@
...
+    # Validate content dialog
+    assert app_dialog.root.cells[
+        selection_hierarchy[1]
+    ].exists(), "{0} is missing in content dialog".format(selection_hierarchy[1])
+    assert app_dialog.root.cells[
+        "{0} - {0}".format(selection_hierarchy[2])
+    ].exists(), "{0} task is missing in content dialog".format(selection_hierarchy[2])
+    assert app_dialog.root.cells[
+        "{0} - {0}".format(entities[3])
+    ].exists(), "{0} task is missing in content dialog".format(entities[3])
```
pscadding, line 111: It looks like these checks are not navigating or validating the breadcrumb so perhaps they should be in a different method?

tests/workfiles2_functions.py
```diff
@@ -0,0 +1,196 @@
...
+    ].exists(), "{0} task should be visible".format(entities[3])
+    assert (
+        app_dialog.root.outlineitems[entities[2]].exists() is False
+    ), "{0} task shouldn't be visible".format(entities[2])
+
+    # Enable My Tasks Only and make sure tasks are not showing up anymore
+    app_dialog.root.checkboxes["My Tasks Only"].mouseClick()
+    assert (
+        app_dialog.root.outlineitems[entities[3]].exists() is False
+    ), "{0} task shouldn't be visible".format(entities[3])
```
pscadding, line 196: A few things here

As mentioned above the entities parameter is a bit vague maybe that needs to be broken up a bit into separate parameters. I didn't break up the selection hierarchy parameter when I suggested it to you, since the order is important. But I'm not sure we are relying on the order of the entities parameter here, it just looks like we are bundling a bunch of entity names together under a single parameter.

Also not a requirement for merging, but it would be nice if this could be broken up into a few smaller methods if possible and you have time. But only do it if it makes sense to do it.

pscadding (approved):
It's looking a lot neater, nice work!

I have made some other suggestions (sorry!) but I wouldn't say they are required, so feel free to ignore if you just need to move on from this.

PR #97, opened 2020-10-05
Add parameter to opening dialogs to specify modal or not.

staceyoue:
@jfboismenu we'd like to add a parameter to the show dialog methods in the Workfiles2 app so that we can choose whether or not to open these dialogs in modal mode. There's a situation that came up for VRED when we need the Workfiles2 save dialog to be modal.


jfboismenu (approved):
Sorry for forgetting to approve this!

PR #98, opened 2020-10-07
SG-15693 add change context mode

pscadding:
Adds a context change mode that means the open and save app options are replaced by a context change app.
In context change mode it contains the entity/context navigation windows, but no file select. Once a context has been picked you can press change context.

Tasks in the My tasks can be view can be double clicked to change context, when in context change mode.

**Refactoring of the Open, Save and Context change forms**

Previously, the open and save dialog derive from the same `FileFormBase` class, but that class makes no assumption about the UI, because the save and open dialogs have different UIs.
However this meant a bunch of duplicated code in the derived classes, because both UI's shared common functionality.
Instead of opting to make context change a "mode" of the open dialog, I decided to make it it's own class derived from the `FileFormBase` and to cut down on the duplication across all three classes I moved a bunch of methods over to the base class. However this means that the base class now has some expectation of what will be in the UI.
I haven't but I could make the ui elements required by the base class, properties which the deriving classes must implement, although the element names would be identical in all cases so I felt this was overkill. Or I could revert to duplicating up the code again across all three classes.

### DIFF
--- a/app.py
+++ b/app.py
@@ -32,11 +32,6 @@ class MultiWorkFiles(sgtk.platform.Application):
             )
             return
 
-        if self.engine.name == "tk-mari":
-            # Mari doesn't have the concept of a current scene so this app shouldn't
-            # provide any commands!
-            return
-
         if self.get_setting("show_change_context"):
             # This will only show the context change dialog and not register the save of open dialogs.
             self.engine.register_command(
--- a/app.py
+++ b/app.py
@@ -56,6 +56,24 @@ class MultiWorkFiles(sgtk.platform.Application):
             },
         )
 
+        self.engine.register_command(
+            "Change Context...",
+            self.show_context_change_dlg,
+            {
+                "short_name": "change_context",
+                # dark themed icon for engines that recognize this format
+                "icons": {
+                    "dark": {
+                        "png": os.path.join(
+                            os.path.dirname(__file__),
+                            "resources",
+                            "file_open_menu_icon.png",
+                        )
+                    }
+                },
+            },
+        )
+
         # register the file save command
         self.engine.register_command(
             "File Save...",
@@ -114,6 +132,12 @@ class MultiWorkFiles(sgtk.platform.Application):
         Launch the main File Open UI
         """
         self._tk_multi_workfiles.WorkFiles.show_file_open_dlg()
+
+    def show_context_change_dlg(self):
+        """
+        Launch the main File Open UI
+        """
+        self._tk_multi_workfiles.WorkFiles.show_context_change_dlg()
 
     def show_file_save_dlg(self):
         """
--- a/app.py
+++ b/app.py
@@ -37,7 +37,7 @@ class MultiWorkFiles(sgtk.platform.Application):
             # provide any commands!
             return
 
-        if self.get_setting("context_change_only_mode"):
+        if self.get_setting("show_change_context"):
             # This will only show the context change dialog and not register the save of open dialogs.
             self.engine.register_command(
                 "Change Context...",
@@ -56,7 +56,7 @@ class MultiWorkFiles(sgtk.platform.Application):
                     },
                 },
             )
-        else:
+        if self.get_setting("show_file_open"):
             # This show the open and save dialogs and not the context change dialog.
             # register the file open command
             self.engine.register_command(
@@ -77,6 +77,7 @@ class MultiWorkFiles(sgtk.platform.Application):
                 },
             )
 
+        if self.get_setting("show_file_save"):
             # register the file save command
             self.engine.register_command(
                 "File Save...",
--- a/app.py
+++ b/app.py
@@ -56,7 +56,7 @@ class MultiWorkFiles(sgtk.platform.Application):
                     },
                 },
             )
-            # else:
+        else:
             # This show the open and save dialogs and not the context change dialog.
             # register the file open command
             self.engine.register_command(
--- a/app.py
+++ b/app.py
@@ -37,61 +37,64 @@ class MultiWorkFiles(sgtk.platform.Application):
             # provide any commands!
             return
 
-        # register the file open command
-        self.engine.register_command(
-            "File Open...",
-            self.show_file_open_dlg,
-            {
-                "short_name": "file_open",
-                # dark themed icon for engines that recognize this format
-                "icons": {
-                    "dark": {
-                        "png": os.path.join(
-                            os.path.dirname(__file__),
-                            "resources",
-                            "file_open_menu_icon.png",
-                        )
-                    }
+        if self.get_setting("context_change_only_mode"):
+            # This will only show the context change dialog and not register the save of open dialogs.
+            self.engine.register_command(
+                "Change Context...",
+                self.show_context_change_dlg,
+                {
+                    "short_name": "change_context",
+                    # dark themed icon for engines that recognize this format
+                    "icons": {
+                        "dark": {
+                            "png": os.path.join(
+                                os.path.dirname(__file__),
+                                "resources",
+                                "file_open_menu_icon.png",
+                            )
+                        }
+                    },
                 },
-            },
-        )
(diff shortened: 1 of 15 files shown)


app.py
```diff
@@ -37,43 +37,64 @@ def init_app(self):
...
-                # dark themed icon for engines that recognize this format
-                "icons": {
-                    "dark": {
-                        "png": os.path.join(
-                            os.path.dirname(__file__),
-                            "resources",
-                            "file_open_menu_icon.png",
-                        )
-                    }
+        if self.get_setting("context_change_only_mode"):
```
jfboismenu, line 40: I think a more usefull approach for clients would be to have the following setting and their defaults

```
show_file_open: true
show_file_save: true
shot_change_context: false
```

This way we allow clients to select what they want. What do you think?

python/tk_multi_workfiles/actions/context_change_action.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2015 Shotgun Software Inc.
```
jfboismenu, line 1:
```suggestion
# Copyright (c) 2020 Shotgun Software Inc.
```

python/tk_multi_workfiles/actions/context_change_action.py
```diff
@@ -0,0 +1,67 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Action to create a new file.
```
jfboismenu, line 12:
```suggestion
Action to create a change context.
```

python/tk_multi_workfiles/actions/context_change_action.py
```diff
@@ -0,0 +1,67 @@
...
+        Constructor.
+
+        :param environment: A WorkArea instance containing the context we will switch to.
+        """
+        Action.__init__(self, "Change Context")
+        self._environment = environment
+
+    def execute(self, parent_ui):
+        """
+        Perform a new-scene operation initialized with the current context.
```
jfboismenu, line 38:
```suggestion
        Perform a context change operation.
```

python/tk_multi_workfiles/actions/context_change_action.py
```diff
@@ -0,0 +1,67 @@
...
+                self._app.log_exception("Unable to run folder creation!")
+
+            if not self._environment.context == self._app.context:
+                # Change context
+                FileAction.change_context(self._environment.context)
+
+        except Exception as e:
+            error_title = "Failed to complete '%s' action" % self.label
+            QtGui.QMessageBox.information(
+                parent_ui, "%s" % error_title, "%s:\n\n%s" % (error_title, e)
```
jfboismenu, line 62:
```suggestion
                parent_ui, error_title, "%s:\n\n%s" % (error_title, e)
```
No need for string interpolation if we only use error_title as is 😉

python/tk_multi_workfiles/work_files.py
```diff
@@ -102,7 +102,17 @@ def show_file_open_dlg():
         handler = WorkFiles()
         from .file_open_form import FileOpenForm
 
-        handler._show_file_dlg("File Open", FileOpenForm)
+        handler._show_file_dlg("", FileOpenForm)
```
jfboismenu, line 105: Is that a mistake? It doesn't say File Open anymore.

jfboismenu (changes requested):
Nice job! This is much cleaner that I expected it would be. I like that we were able to refactor more code to the base class. What you did wasn't unreasonable. Each dialog have the same general layout, so it makes sense for some of the callbacks to be shared.

There's one likely bug in there however, I think File Open does not have a title anymore.

I assume you've carefully went through the remove code in file_open_form and file_save_form to make sure the callbacks were exactly the same?

Awesome work!

app.py
pscadding: I don't really, mind, I guess it allows more flexibly. I'm not sure when I would have the open app and not the save app, I guess maybe in a project context

python/tk_multi_workfiles/actions/context_change_action.py
pscadding: thats annoying I did a pass on this file, to make sure I'd renamed it. It was obviously a bad pass :(

python/tk_multi_workfiles/actions/context_change_action.py
pscadding: sure! I think I just copied it from somewhere else in the code, so that should be changed as well, I'll have a look.

python/tk_multi_workfiles/work_files.py
pscadding: I just wanted to mess with people. I have no idea how that crept in, thanks!

pscadding:
I've made a few small changes to get it to work with Mari, you can see the diff here:
https://github.com/shotgunsoftware/tk-multi-workfiles2/compare/ce04f21..9729fe6

python/tk_multi_workfiles/actions/context_change_action.py
```diff
@@ -0,0 +1,65 @@
...
+        Perform a context change operation.
+
+        :param parent_ui: Parent dialog executing this action.
+        """
+
+        try:
+            # create folders and validate that we can save using the work template:
+            try:
+                # create folders if needed:
+                FileAction.create_folders(self._environment.context)
```
jfboismenu, line 47: Why the change from `create_folders_if_needed` to `create_folders`? I'm looking at the code and I understand the intent, but I don't understand it in this context (no pun intended 😜).

app.py
```diff
@@ -32,48 +32,65 @@ def init_app(self):
             )
             return
 
-        if self.engine.name == "tk-mari":
```
jfboismenu, line 35: Sounds like asking to actually test in mari paid off. 😃

hooks/scene_operation_tk-mari.py
```diff
@@ -0,0 +1,67 @@
...
+
+        :returns:               Depends on operation:
+                                'current_path' - Return the current scene
+                                                 file path as a String
+                                'reset'        - True if scene was reset to an empty
+                                                 state, otherwise False
+                                all others     - None
+        """
+
+        # Mari doesn't have any scene operations.
```
jfboismenu, line 66: Maybe we could elaborate a bit more here, explaining that the Mari workflow is based on context switching instead of opening scene files?

jfboismenu (commented):
I'm unsure about some of your changes. Could you clarify the comments and explain why the change from `create_folders_if_needed` to `create_folders`?

python/tk_multi_workfiles/actions/context_change_action.py
pscadding: If you don't actually have a template then the `create_folders_if_needed` method will error since it expects a template. With Mari you won't have a template, and if you are changing context even if you set it up in Maya, it doesn't necessarily make sense to expect a template either.

As far as I can gather from the `create_folders_if_needed` method, it is there to try and avoid folder creation if possible based on resolving templates. Since we can't rely on that, creating the folders without the check seems like the better option.

I had initially started to modify that method to check if there was a template, but I thought it probably just makes sense to call the `create_folders` instead.

pscadding: I guess going back to using `create_folders_if_needed` and then modifying it, would mean that if you say were using it in Maya where a template did exist, you could avoid the folder creation overhead, potentially. 
I'm happy to go with either approach.

jfboismenu (approved):
Everything is awesome! Thank you!

PR #100, opened 2020-11-13
file_save_form and file_open_form now send through the correct parent…

mark-oshea:
… action to _get_current_file() in file_form_base


mark-oshea:
The reason for this is because in scene_operation hooks, I can't differentiate between whether the 'current_file' operation is for a File Open or File Save.. I've been writing some custom behaviours for Nukestudio scene operation and want it to behave in a specific way if it receives a 'current_file' operation (but only if the parent action is Open File)

PR #101, opened 2020-12-01
SG-20264 Support Email Update

eshokrgozar:
Switched to using support_url from tk-core instead of the hard-coded email address. Also using TKMessageBox from tk-core to make sure hyperlinks are clickable and styled


python/tk_multi_workfiles/ui/file_save_form.py
```diff
@@ -166,7 +166,7 @@ def setupUi(self, FileSaveForm):
         self.verticalLayout_4.setObjectName("verticalLayout_4")
         self.preview_grid = QtGui.QGridLayout()
         self.preview_grid.setHorizontalSpacing(14)
-        self.preview_grid.setVerticalSpacing(6)
+        self.preview_grid.setVerticalSpacing(1)
```
pscadding, line 169: I'm guessing this is a result of modifying the UI file and recompiling but isn't an intentional change. Does it make any visual difference?

pscadding (approved):
The tests seem unhappy I guess this is because they don't have the updated core?

python/tk_multi_workfiles/ui/file_save_form.py
eshokrgozar: There was an issue with the warning message being clipped now that it is longer. This seems to solve the problem. I'd love to come up with a better solution but I couldn't figure out the root of the problem

PR #102, opened 2020-12-17
Adding Azure Agent Name to Shotgun project name

mathurf:
Adding Azure Agent Name to Shotgun project name.
Like that, the Shotgun project will not be overwritten when we will run the UI automation on both, Python 2 and 3

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,7 +14,7 @@ resources:
     - repository: templates
       type: github
       name: shotgunsoftware/tk-ci-tools
-      ref: refs/heads/mathurf_hackathon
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -42,6 +42,7 @@ jobs:
 - template: build-pipeline.yml@templates
   parameters:
     has_ui_tests: true
+    tk_toolchain_ref: mathurf_hackthon
```
jfboismenu, line 45: Don't forget to take this line out! 😄

jfboismenu (approved):
image

PR #104, opened 2021-02-01
SHOT-3645: Fix New Task dialog to only pop up once.

staceyoue:
* New Task dialog will pop up a second time after creating a new task due to duplicated signal connection.
* Removed signal connection in `FileOpenForm` as it is already connected in the parent class `FileFormBase`


thebeeland (approved):
Looks good!

PR #105, opened 2021-02-02
SHOT-3474: Include extra details on "New File" action error

staceyoue:
If an Asset has no type set, this error being reported to the user in the UI did not include the more accurate error message that was showing up in the log file:

improve-error-flow

With this change, the dialog includes an extra details message (if there are any details) that are shown when clicking "Show Details":

Screen Shot 2021-02-02 at 5 23 29 PM

See JIRA ticket for more detail: https://jira.autodesk.com/browse/SHOT-3474

### DIFF
--- a/python/tk_multi_workfiles/actions/new_file_action.py
+++ b/python/tk_multi_workfiles/actions/new_file_action.py
@@ -70,35 +70,54 @@ class NewFileAction(Action):
                 self._environment.context.as_template_fields(
                     self._environment.work_template, validate=True
                 )
-            except TankError as e:
-                # log the original exception (useful for tracking down the problem)
+            except TankError as tank_error:
                 self._app.log_exception(
                     "Unable to resolve template fields after folder creation!"
                 )
-                # and raise a new, clearer exception for this specific use case:
-                raise TankError(
-                    "Unable to resolve template fields after folder creation!  This could mean "
-                    "there is a mismatch between your folder schema and templates. Please "
-                    "contact us via {} if you need help fixing this.".format(
-                        support_url
-                    )
-                )
 
-            # reset the current scene:
-            if not reset_current_scene(
-                self._app, NEW_FILE_ACTION, self._environment.context
-            ):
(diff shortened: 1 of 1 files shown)


python/tk_multi_workfiles/actions/new_file_action.py
```diff
@@ -79,8 +79,8 @@ def execute(self, parent_ui):
                 raise TankError(
                     "Unable to resolve template fields after folder creation!  This could mean "
                     "there is a mismatch between your folder schema and templates. Please "
-                    "contact us via {} if you need help fixing this.".format(
-                        support_url
+                    "contact us via {url} if you need help fixing this.\n\n{error}".format(
+                        url=support_url, error=str(e)
                     )
                 )
```
thebeeland, line 86: Rather than building a string, raising it, catching it, parsing it, and then displaying it, I'd like to see this as a try/except/else block of logic for this specific error case, and then the previous logic with the `QtGui.QMessageBox.information` for the general exception handling farther down.

So, up here, you'll have something like:

```python
try:
    FileAction.create_folders...
except TankError as e:
    self._app.log_exception...
    msg_box = QtGui.QMessageBox( ... )
    msg_box.setDefaultButton...
    msg_box.exec_()
    return False
else:
    ...
```

You'll just skip this raise entirely by handling the exception you got here and returning. Then you don't need the logic below that's trying to dissect the exception text and present it nicely.

thebeeland (changes requested):
I like the idea, but I'd like to see the solution approached a little differently just to reduce the amount of string manipulation involved.

python/tk_multi_workfiles/actions/new_file_action.py
staceyoue: Agreed, this way is much simpler -- fixed.

thebeeland: Thanks! I'll look it over again this afternoon.

thebeeland (approved):
Perfection!

PR #110, opened 2021-11-24
SG-21170-ROUND2 [ZD to AKN] Update links referenced from code to the old support site to point to AKN

000paradox000:
Update URLs for documentation and toolkit general information


mathurf (approved):
LGTM!

PR #112, opened 2021-11-30
SG-20204 Remove Softimage Support on February 13th

NorberMV:
Remove scene_operation_tk-softimage.py hook from app, because it will not be supported anymore.


mathurf (approved):
LGTM!

PR #118, opened 2022-07-11
[SHOT-4065] Check references on file open

staceyoue:
* Add checkbox option next to "All Versions" to indicate if references should be checked on file open
* Only workfiles will be checked (not published), even if the option is toggled on
* If out of date references are found, a pop up dialog will be shown to ask the user how to handle them


000paradox000 (approved):
LGTM

PR #120, opened 2022-11-10
SG-25609 Nuke 13 Crashes On Saving With a Custom Root Path

NorberMV:
This PR  fixes the issue with Nuke 13 on Windows failing to save when we have path definitions in the template starting with `0`, e.g.:
` shot_root: 08 sequences/{Sequence}/{Shot}/{Step`

### DIFF
--- a/hooks/scene_operation_tk-nuke.py
+++ b/hooks/scene_operation_tk-nuke.py
@@ -102,7 +102,7 @@ class SceneOperation(HookClass):
 
         if operation == "current_path":
             # return the current script path
-            return nuke.root().name().replace("/", os.path.sep)
+            return nuke.root().name().replace(os.path.sep, "/")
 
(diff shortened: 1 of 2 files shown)


000paradox000 (approved):
LGTM nice work

badgrenola:
Hello! 

It looks like not all of the changes from this PR made it into _master_. I can only see part of the `f00d35df3083e7ba0bdcc04571774768578f38f2` commit. 

image

I just wanted to check if that was intentional, as we're seeing the same crash. 

Cheers!

carlos-villavicencio-adsk:
Hey @badgrenola, yes it was intentional. The changes that appear in other commits are rolled back in the following ones.

Regarding the crash, we're working on fixing this once #126 is merged/released.

PR #121, opened 2022-12-09
SG-29395: Ensure that the overlay is hidden on shut down.

staceyoue:
* Without hiding the widget before shut down, the overlay spinner continues to fire timeout events to trigger animations.


NorberMV (approved):
LGTM!

000paradox000 (approved):

PR #122, opened 2022-12-29
SG 23213 Opening Nuke file with wrong minor version fails

NorberMV:
This PR catch and handles the exception raised when an user open a Nuke file with a different minor version, this just raise a QMessageBox warning but doesn't restore the original context as Nuke doesn't  take this as a failed operation, so  keep the context that the work file should be opened in.

### DIFF
--- a/python/tk_multi_workfiles/actions/open_file_action.py
+++ b/python/tk_multi_workfiles/actions/open_file_action.py
@@ -170,7 +170,10 @@ class OpenFileAction(FileAction):
             # warning but doesn't restore the original context as Nuke doesn't
             # take this as a failed operation, so let's keep the context that the
             # work file should be opened in.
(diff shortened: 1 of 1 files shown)


000paradox000 (approved):
Great job!

python/tk_multi_workfiles/actions/open_file_action.py
```diff
@@ -210,7 +210,7 @@ def __handle_open_file_exception(
         # take this as a failed operation, so let's keep the context that the
         # work file should be opened in.
         if isinstance(exc_value, RuntimeError) and re.match(
-            r"x\S+.v[\d.]+.nk is for nuke[\d.]+v[\d.]+; this is nuke[\d.]+v[\d.]+",
+            r"\S+.v[\d.]+.nk is for nuke[\d.]+v[\d.]+; this is nuke[\d.]+v[\d.]+",
```
NorberMV, line 213: Nice catch! thanks @vazquea

PR #123, opened 2023-01-18
SG-29639: Improve search filter

staceyoue:
* Apply search text filter once user finishes editing, instead of on each keystroke.


NorberMV (approved):
LGTM!, so this change would require [this](https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/135) modification in tk-framework-qtwidgets  as a required "minimum_version", right ?

staceyoue:
> LGTM!, so this change would require [this](https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/135) modification in tk-framework-qtwidgets as a required "minimum_version", right ?

Yes, it is required for this improvement to take full effect, but if the minimum version is not updated it will not break the app.

PR #126, opened 2023-02-16
SG-29904 Update string replace to be compatible with Nuke on win32

carlos-villavicencio-adsk:
- Follow up from #120.
- Detected when Nuke 13.2v5 crashes when saving in ShotGrid
- Nuke uses a `/` as a separator on all platforms (including Windows)


NorberMV (approved):
LGTM !, looks like we're going to have to track these replaces on all hooks using the Nuke Python API.

julien-lang (approved):
LGTM!

PR #127, opened 2023-03-10
SG-30400 Retrieve thumbnails from any FileItem that has a 'thumbnail_path'

kennedy0:
The filter_work_files hook allows you to provide a thumbnail, but it's not used due to the `file_item.is_published` requirement in the FileModel._process_files method. This leaves us unable to show thumbnails for working files.

With this requirement removed, the filter_work_files hook can provide thumbnails as a file URI:

```python
class FilterWorkFiles(HookClass):
    def execute(self, work_files, **kwargs):
        for wf in work_files:
            wf['work_file']['thumbnail'] = "file:///my_thumbnail.jpg"
        return work_files
```

### DIFF
--- a/python/tk_multi_workfiles/file_model.py
+++ b/python/tk_multi_workfiles/file_model.py
@@ -842,11 +842,9 @@ class FileModel(QtGui.QStandardItemModel):
             # add to the list of valid files:
             valid_files[file_version_key] = file_item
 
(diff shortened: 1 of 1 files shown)


carlos-villavicencio-adsk:
Thank you for this contribution. We will review this internally before merging.

PR #129, opened 2023-04-04
SG-30567 Change string interpolation syntax

carlos-villavicencio-adsk:
To support DCCs running on Python 2 (Nuke 12).

Follow up from #124

### DIFF
--- a/python/tk_multi_workfiles/user_cache.py
+++ b/python/tk_multi_workfiles/user_cache.py
@@ -152,7 +152,7 @@ class UserCache(Threaded):
                     "id": None,
                     "email": None,
                     "login": login_name,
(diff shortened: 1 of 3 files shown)


python/tk_multi_workfiles/user_cache.py
```diff
@@ -152,7 +152,7 @@ def _get_user_details_for_login(self, login_name):
                     "id": None,
                     "email": None,
                     "login": login_name,
-                    "name": f"{login_name} (System)",
+                    "name": "%s (System)" % login_name,
```
julien-lang, line 155: You can still use the format function:
```python
"{login_name} (System)".format(login_name=login_name)
```

julien-lang (approved):

PR #130, opened 2023-06-12
SHOT-4259: Remove tk-alias engine to run at startup

staceyoue:
* tk-alias engine will be responsible for running the workfiles app at start up using the config setting `run_at_startup`
* tk-vred will also be removed from the workfiles2 supported engines list, to be consistent with tk-alias/Automotive apps


rob-aitchison:
@staceyoue & @barbara-darkshot , should we also drop VRED here for consistency sake?

rob-aitchison (approved):
Thanks for making our Engines the same - cheers!

PR #132, opened 2023-08-04
SG-31598: List view

staceyoue:
Reviving this old branch to add a list view to the app.

* Add a list view mode to the file view
* Uses the ViewItemDelegate from tk-framework-qtwidgets

~~Requires this PR: https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/142~~
Minimum tk-framework-qtwidgets version bumped to v2.10.8


python/tk_multi_workfiles/my_tasks/my_tasks_model.py
```diff
@@ -13,19 +13,27 @@
...
+class MyTasksModel(ShotgunExtendedEntityModel, ViewItemRolesMixin):
     """
     Specialisation of the Shotgun entity model that represents a single users tasks.  Note that we derive
     from the Shotgun entity model so that we have access to the entity icons it provides.  These are used
-    later by the MyTaskItemDelegate when rending a widget for a task in the My Tasks view.
+    later by the delegate when rendering a task item in the My Tasks view.
     """
 
+    _BASE_ROLE = QtCore.Qt.UserRole + 32
+    (NEXT_AVAILABLE_ROLE,) = range(_BASE_ROLE, _BASE_ROLE + 1)
```
carlos-villavicencio-adsk, line 33: What does 32 mean? Is it possible to extract this magic value to a constant?

staceyoue: This is a Toolkit legacy thing...

Basically the purpose here is to ensure we use model item data roles that Qt does not use. Qt provides `QtCore.Qt.UserRole` as the first model item data role that applications can use to add their own data roles. Most data models in Toolkit derive from `ShotgunModel`, which derives from the `ShotgunQueryModel`; both of these models define some custom roles (for example [here](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/57935971acd686210db540b5bd8ef842b373d321/python/shotgun_model/shotgun_query_model.py#L102-L108)). So I believe the arbitrary number of "32" was picked to ensure any model deriving from these models, would not override the base model custom roles. While this method works, it is not a bullet proof since if the base models needed to define more custom roles, and exceeded 32, then we would have conflicts.

So all that to say, here I am following the precedent set by other Toolkit apps, but it would be ideal if the shotgun models in `tk-framework-shotgunutils` were modified to provide a more "official" model role that is safe to use by models that inherit the base shotgun models (similarly to how Qt provides `UserRole`).

I am happy to implement the proposed change, but it extends the scope of this ticket. Do we want to include this, or possibly create a new ticket for it?

resources/file_list_form.ui
```diff
@@ -30,60 +30,180 @@
...
-    <layout class="QHBoxLayout" name="horizontalLayout_3" stretch="0,0,0,1,0">
+    <layout class="QHBoxLayout" name="horizontalLayout_8">
+    <!-- <layout class="QHBoxLayout" name="horizontalLayout_3" stretch="0,0,0,1,0">
      <property name="leftMargin">
       <number>1</number>
      </property>
      <property name="rightMargin">
       <number>1</number>
-     </property>
+     </property> -->
```
julien-lang, line 40: Do we want to keep this comment for backup?
Was this change made by QTDesigner or manually?

julien-lang (approved):
LGTM!

Just a minor comment about XML comments.

resources/file_list_form.ui
staceyoue: Good catch, I think this was a leftover merge conflict I was deciding between. It has been removed.

hooks/ui_config_tk-houdini.py
```diff
@@ -0,0 +1,33 @@
...
+        :type item: :class:`sgkt.platofrm.qt.QtGui.QStandardItem`
+
+        :return: The QBrush.
+        :rtype: :class:`sgtk.platform.qt.QtGui.QBrush`
+        """
+
+        parent = item.index().parent()
+        if parent.isValid():
+            return QtGui.QApplication.palette().base()
+        return None
```
julien-lang, line 33: Useless line. If there is no return, a Python function returns None.

But I'd prefer a fail-first strategy instead (easier to maintain IMO):
```python
parent = item.index().parent()
if not parent.isValid():
    return

return QtGui.QApplication.palette().base()
```

Not a blocker. Just my preference :)

staceyoue: Returning `None` here effectively says "use the default brush". When it is time to draw, Qt will use the preferred default brush when the brush is None.

I can remove the `return None` line though.

PR #141, opened 2024-01-16
SHOT-4267: Use ShotgunMenu instead of QMenu

staceyoue:
* For style consistency across apps
* Fixes QMenu right-arrow display issue with PySide6


carlos-villavicencio-adsk (approved):
💯

PR #147, opened 2024-05-01
SG-34684 Add task status filter dropdown

carlos-villavicencio-adsk:
image


python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +675,50 @@ def _on_step_filter_changed(self, step_list):
...
+        if value == "ALL":
+            self.my_tasks_model.load_and_refresh()
+        else:
+            self.my_tasks_model.load_and_refresh(
+                extra_filter=["sg_status_list", "is", value.lower()]
+            )
+
+        self.my_tasks_model.data_refreshed.emit(True)
+        self.my_tasks_model.async_refresh()
+        return
```
julien-lang, line 724: useless return since it's the last instruction of the function.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +675,50 @@ def _on_step_filter_changed(self, step_list):
...
+
+    # Task status filter methods
+    def __get_project_id(self, bundle):
+        """
+        Helper method to return the project id for the current context.
+
+        :param bundle: The current app bundle.
+        """
+        if bundle.tank.pipeline_configuration.is_site_configuration():
+            return None
```
julien-lang, line 687: Useless None value. Return without args raises none.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +675,50 @@ def _on_step_filter_changed(self, step_list):
...
+        else:
+            return bundle.tank.pipeline_configuration.get_project_id()
```
julien-lang, line 689: Minor: no need for a "else" statement. Embrace fail-first and avoid right-indent.

python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +675,50 @@ def _on_step_filter_changed(self, step_list):
...
+        app = sgtk.platform.current_bundle()
+        project_id = self.__get_project_id(app)
+        if project_id:
+            sg = app.shotgun._sg
+            logger.debug(
+                f"Task Status Filter: getting statuses from project {project_id}"
+            )
+            task_status_list = sg.schema_field_read(
+                "Task", "sg_status_list", {"type": "Project", "id": project_id}
+            )["sg_status_list"]["properties"]["valid_values"]["value"]
```
julien-lang, line 704: I would protect that by a try/except KeyError because we rely too much on the received data structure.

resources/entity_tree_form.ui
```diff
@@ -40,6 +40,13 @@
      <property name="bottomMargin">
       <number>1</number>
      </property>
+     <item>
+      <widget class="QComboBox" name="task_status_combo">
+       <property name="text">
+        <string>Task Status</string>
```
julien-lang, line 46: _Task Status_ or _Task**s** Status_ ?

python/tk_multi_workfiles/ui/entity_tree_form.py
```diff
@@ -19,6 +19,9 @@ def setupUi(self, EntityTreeForm):
         self.horizontalLayout = QtGui.QHBoxLayout()
         self.horizontalLayout.setContentsMargins(2, -1, 2, 1)
         self.horizontalLayout.setObjectName("horizontalLayout")
+        self.task_status_combo = QtGui.QComboBox(EntityTreeForm)
+        self.task_status_combo.setObjectName("task_status_combo")
+        self.horizontalLayout.addWidget(self.task_status_combo)
```
julien-lang, line 24: What's it's default visibility? I'd vote for default hidden

python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +675,50 @@ def _on_step_filter_changed(self, step_list):
...
+        Populate the task status list from the project schema.
+        """
+        app = sgtk.platform.current_bundle()
+        project_id = self.__get_project_id(app)
+        if project_id:
+            sg = app.shotgun._sg
+            logger.debug(
+                f"Task Status Filter: getting statuses from project {project_id}"
+            )
+            task_status_list = sg.schema_field_read(
```
staceyoue, line 702: You should be able to use the app `shotgun`  property directly to make api requests:
```suggestion
            task_status_list = app.shotgun.schema_field_read(
```

staceyoue (changes requested):

resources/entity_tree_form.ui
carlos-villavicencio-adsk: I don't have a strong opinion. What should be the best?

python/tk_multi_workfiles/ui/entity_tree_form.py
carlos-villavicencio-adsk: How can I achieve a default visibility in the UI form? Do you have an example?

julien-lang: Not sure it will work but I think of this one: https://github.com/shotgunsoftware/tk-core/blob/v0.21.2/python/tank/authentication/login_dialog.py#L248

python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +675,40 @@ def _on_step_filter_changed(self, step_list):
...
+        if project_id:
+            logger.debug(
+                f"Task Status Filter: getting statuses from project {project_id}"
+            )
+            try:
+                task_status_list = app.shotgun._sg.schema_field_read(
+                    "Task", "sg_status_list", {"type": "Project", "id": project_id}
+                )["sg_status_list"]["properties"]["valid_values"]["value"]
+            except KeyError:
+                task_status_list = []
```
julien-lang, line 695: Do we need to populate a single "ALL" entry in this case? That does not make sense. It'd be better to hide/grey the widget in this case

python/tk_multi_workfiles/ui/entity_tree_form.py
carlos-villavicencio-adsk: The default state is visible.

julien-lang (approved):
LGTM!

carlos-villavicencio-adsk:
@staceyoue can you take another look, please?

python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +679,37 @@ def _on_step_filter_changed(self, step_list):
...
+        """
+        Populate the task status list from the project schema.
+        """
+        app = sgtk.platform.current_bundle()
+        project_id = app.context.project.get("id")
+        if project_id:
+            logger.debug(
+                f"Task Status Filter: getting statuses from project {project_id}"
+            )
+            task_status_list = app.shotgun._sg.schema_field_read(
```
staceyoue, line 694: You should not need to access the `_sg` property, just use `shotgun` itself
```suggestion
            task_status_list = app.shotgun.schema_field_read(
```

carlos-villavicencio-adsk: Didn't work to me. 

I see the [shotgun](https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/master/app.py#L166) property returns a `DebugWrapperShotgun`. From there I can access the `_sg` [instance](https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/master/app.py#L189) as property from it.

staceyoue (approved):

python/tk_multi_workfiles/browser_form.py
```diff
@@ -669,3 +679,37 @@ def _on_step_filter_changed(self, step_list):
...
+        task_status_list.insert(0, "ALL")
+        self._my_tasks_form._ui.task_status_combo.addItems(task_status_list)
+
+    def _on_task_status_combo_changed(self, value):
+        """
+        Called when the task status combo box is changed.
+
+        :param value: The selected value on the combo box.
+        """
+        if value == "ALL":
```
eduardoChaucaGallegos, line 707: "ALL" string is used in 2 functions of same class so maybe we can use a new variable in the class scope to be used in this condition and in a previous insert(), because if that string change in future it will need to be changed 2 times

PR #150, opened 2024-07-31
SG-33500 New build resources yml

eduardoChaucaGallegos:
- New build resources yml

PR #151, opened 2024-10-08
SG-36186 Fix ui sources

carlos-villavicencio-adsk:
Fix ui sources from #147 and #133

### DIFF
--- a/python/tk_multi_workfiles/ui/entity_tree_form.py
+++ b/python/tk_multi_workfiles/ui/entity_tree_form.py
@@ -3,7 +3,7 @@
 ################################################################################
 ## Form generated from reading UI file 'entity_tree_form.ui'
 ##
-## Created by: Qt User Interface Compiler version 5.15.2
+## Created by: Qt User Interface Compiler version 5.15.16
 ##
 ## WARNING! All changes made in this file will be lost when recompiling UI file!
 ################################################################################
--- a/resources/entity_tree_form.ui
+++ b/resources/entity_tree_form.ui
@@ -42,9 +42,6 @@
      </property>
      <item>
       <widget class="QComboBox" name="task_status_combo">
-       <property name="text">
-        <string>Task Status</string>
-       </property>
       </widget>
      </item>
      <item>
--- a/python/tk_multi_workfiles/ui/file_list_form.py
+++ b/python/tk_multi_workfiles/ui/file_list_form.py
@@ -3,7 +3,7 @@
 ################################################################################
 ## Form generated from reading UI file 'file_list_form.ui'
 ##
-## Created by: Qt User Interface Compiler version 5.15.2
+## Created by: Qt User Interface Compiler version 5.15.16
 ##
 ## WARNING! All changes made in this file will be lost when recompiling UI file!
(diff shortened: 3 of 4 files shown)


eduardoChaucaGallegos (approved):
LGTM, but please review a few checks from the pipeline.

carlos-villavicencio-adsk:
Hey @staceyoue , can you take a look at this? It's a follow up work caused because some generated files were updated in #133.

resources/entity_tree_form.ui
```diff
@@ -42,9 +42,6 @@
      </property>
      <item>
       <widget class="QComboBox" name="task_status_combo">
-       <property name="text">
-        <string>Task Status</string>
-       </property>
```
staceyoue, line 47: Just checking that we meant to remove this property?

staceyoue (approved):
LGTM

python/tk_multi_workfiles/ui/entity_tree_form.py
```diff
@@ -3,7 +3,7 @@
 ################################################################################
 ## Form generated from reading UI file 'entity_tree_form.ui'
 ##
-## Created by: Qt User Interface Compiler version 5.15.2
+## Created by: Qt User Interface Compiler version 5.15.16
```
julien-lang, line 6: Should we freeze a specific version to the "patch" version?

julien-lang (approved):

resources/entity_tree_form.ui
carlos-villavicencio-adsk: This PR intends to leave the generated python UI files as #133. 

Since you were the author of that PR, let us know if we can remove it finally or not.

python/tk_multi_workfiles/ui/entity_tree_form.py
carlos-villavicencio-adsk: 5.15.16 was the version I used to generated the UI sources. It doesn't make any difference in this PR but in the future, I'd suggest we should generate files with PySide versions we support.

resources/entity_tree_form.ui
staceyoue: Hm forgive me, I don't remember removing this or not, since that PR was from last year. Looking at the PR though, there were no modifications to this file `resources/entity_tree_form.ui`

carlos-villavicencio-adsk: I did some historical GIT research and found out that this block was introduced on #147 manually updating the generated files.

I'll do some tests and put back this block of code if needed. Thanks for the heads up

carlos-villavicencio-adsk: Update: this block generates [this line](https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/150/commits/f5d88b2f82b9749e3dbd5efab3db0195aab2184f#diff-d440001cd0e979906cf2f3bf25d9c39c186ef81cb6ecb5a677d113c2ed6cf1b5L104) on `python/tk_multi_workfiles/ui/entity_tree_form.py` which breaks the layout like this

image

The best option is to keep it out.

PR #152, opened 2024-10-17
SG-35979 context cache 

eduardoChaucaGallegos:
- The context cache is saved at the time of the scene operation, specifically during the save process.


julien-lang (commented):
I don't really undertstand what this PR tries to fix...

Also the ticket mentioned that this `_add_to_context_cache` method should also be called in a "finalize" block.

carlos-villavicencio-adsk (approved):

PR #154, opened 2024-11-12
SG-36835 - python2 removal

eduardoChaucaGallegos:
- Following the steps from the plan checklist

### DIFF
--- a/hooks/scene_operation_tk-houdini.py
+++ b/hooks/scene_operation_tk-houdini.py
@@ -13,11 +13,6 @@ import hou
 
 import sgtk
 
-try:
-    from tank_vendor import sgutils
-except ImportError:
-    from tank_vendor import six as sgutils
-
 HookClass = sgtk.get_hook_baseclass()
 
 
@@ -75,13 +70,13 @@ class SceneOperation(HookClass):
         elif operation == "open":
             # give houdini forward slashes
             file_path = file_path.replace(os.path.sep, "/")
-            hou.hipFile.load(sgutils.ensure_str(file_path))
+            hou.hipFile.load(str(file_path))
         elif operation == "save":
             hou.hipFile.save()
         elif operation == "save_as":
             # give houdini forward slashes
             file_path = file_path.replace(os.path.sep, "/")
-            hou.hipFile.save(sgutils.ensure_str(file_path))
+            hou.hipFile.save(str(file_path))
         elif operation == "reset":
             hou.hipFile.clear()
             return True
--- a/hooks/scene_operation_tk-houdini.py
+++ b/hooks/scene_operation_tk-houdini.py
@@ -13,7 +13,10 @@ import hou
 
 import sgtk
 
-from tank_vendor import six
+try:
+    from tank_vendor import sgutils
+except ImportError:
+    from tank_vendor import six as sgutils
 
 HookClass = sgtk.get_hook_baseclass()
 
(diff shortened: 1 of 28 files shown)


hooks/scene_operation_tk-houdini.py
```diff
@@ -72,13 +75,13 @@ def execute(
         elif operation == "open":
             # give houdini forward slashes
             file_path = file_path.replace(os.path.sep, "/")
-            hou.hipFile.load(six.ensure_str(file_path))
+            hou.hipFile.load(sgutils.ensure_str(file_path))
```
carlos-villavicencio-adsk, line 78: I'm wondering if we can use just `str` or no extra function given that `file_path` is already a string (probably).

python/tk_multi_workfiles/file_finder.py
```diff
@@ -280,9 +273,7 @@ def _process_work_files(
 
             # copy common fields from work_file:
             #
-            file_details = dict(
-                [(k, v) for k, v in six.iteritems(work_file) if k != "path"]
-            )
+            file_details = dict([(k, v) for k, v in work_file.items() if k != "path"])
```
carlos-villavicencio-adsk, line 276: Will it work?

```suggestion
            file_details = {k, v for k, v in work_file.items() if k != "path"}
```

Same question for the rest of this file.

python/tk_multi_workfiles/file_search_cache.py
```diff
@@ -115,7 +114,7 @@ def find_file_versions(self, work_area, file_key, clean_only=False):
             return {}
 
         # return a dictionary of version:FileItem entries:
-        return dict([(v, f) for v, f in six.iteritems(file_info.versions)])
+        return dict([(v, f) for v, f in file_info.versions.items()])
```
carlos-villavicencio-adsk, line 117: if `file_info.versions` is already a dictionary, why do we need to map it to a new dictionary?

Will it work?

```suggestion
        return file_info.versions
```

Maybe I'm missing something...

carlos-villavicencio-adsk (approved):
Looks great, good job!

I wrote some open questions, mostly nitpicking. I don't want you to address them right now, we can refactor them in the future if they make things complicated right now. They're not part of the scope as the rest of the changes you did.

python/tk_multi_workfiles/file_search_cache.py
julien-lang: Maybe we want to send a copy of the object instead of a reference?!

python/tk_multi_workfiles/scene_operation.py
```diff
@@ -107,7 +106,7 @@ def get_current_path(app, action, context):
     """
     app.log_debug("Retrieving current scene path...")
     return _do_scene_operation(
-        app, action, context, "current_path", result_types=(six.string_types,)
+        app, action, context, "current_path", result_types=((str,),)
```
julien-lang, line 109: I am not sure about the type we return here?!

Where is this thing used? Is it really expending a list of list of types?

julien-lang (commented):
Mostly good. I just have one pending question.

python/tk_multi_workfiles/scene_operation.py
eduardoChaucaGallegos: fixed, [_do_scene_operation](https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/master/python/tk_multi_workfiles/scene_operation.py#L47) requires only a tuple

carlos-villavicencio-adsk (approved):

julien-lang (approved):
LGTM!

PR #158, opened 2025-05-13
SG-2824 Sorting my tasks

carlos-villavicencio-adsk:
Adds sorting for **My Tasks** left tab. It allows the user to provide custom fields in the `info.yml` file (assuming that they are valid fields to sort by).

image

**Heavily inspired by:**
- https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/147
- https://github.com/shotgunsoftware/tk-multi-shotgunpanel/pull/67
- https://github.com/shotgunsoftware/tk-multi-shotgunpanel/pull/96

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -143,8 +143,8 @@ configuration:
                       *filters* a list of standard Flow Production Tracking API filters, *hierarchy* a list of fields defining
                       the grouping in the sub-tree.
                       The optional *step_filter_on* setting is an Entity type and can be used to restrict
-                      the list of Steps used for filtering to only the Steps for the specified Entity type.
-"
+                      the list of Steps used for filtering to only the Steps for the specified Entity type."
+
         # For backward compatibility reasons we don't use the new deferred queries
         # by default. Using deferred queries, an Asset/Shot setup could be:
         # - caption: Assets (deferred)
@@ -241,6 +241,27 @@ configuration:
         type: bool
         description: Controls whether new tasks can be created from the app.
         default_value: True
+
+    sort_fields:
+        type: dict
+        description: Controls what fields appear for sorting in the UI.
+                     Currently the sort menu only shows on the left tasks tab.
+                     You can add any field that would normally be suitable for sorting a query
+                     by for the tab's given entity type.
+        allows_empty: True
+        default_value:
+            tasks:
+                - { field_code: due_date, display_name: Due Date }
+                - { field_code: start_date, display_name: Start Date }
+                - { field_code: sg_status_list, display_name: Status }
+                - { field_code: step, display_name: Step}
+        values:
+            type: list
+            values:
+                type: dict
+                items:
+                    field_code: { type: str }
+                    display_name: { type: str }
 
     file_browser_tabs:
         type: list
--- a/python/tk_multi_workfiles/browser_form.py
+++ b/python/tk_multi_workfiles/browser_form.py
@@ -23,7 +23,7 @@ from .file_list.file_list_form import FileListForm
 from .file_model import FileModel
 from .util import value_to_str, get_sg_entity_name_field
 from .ui.browser_form import Ui_BrowserForm
-from .framework_qtwidgets import Breadcrumb
+from .framework_qtwidgets import Breadcrumb, shotgun_menus, shotgun_fields
 
 from .file_filters import FileFilters
 from .util import monitor_qobject_lifetime, get_template_user_keys
@@ -242,6 +242,7 @@ class BrowserForm(QtGui.QWidget):
 
             # Task status filter
             self.my_tasks_model = my_tasks_model
+            self._extra_filter = None
             try:
                 self.populate_task_status_list()
                 self._my_tasks_form._ui.task_status_combo.currentTextChanged.connect(
@@ -250,6 +251,9 @@ class BrowserForm(QtGui.QWidget):
             except KeyError as e:
                 logger.warning("Task Status Filter: Error loading: {}".format(e))
                 self._my_tasks_form._ui.task_status_combo.hide()
+
+            # Sorting
+            self._sort_setup()
 
         for caption, step_filter_on, model in entity_models:
             step_entity_filter = None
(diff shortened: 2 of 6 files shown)


info.yml
```diff
@@ -242,6 +242,27 @@ configuration:
         description: Controls whether new tasks can be created from the app.
         default_value: True
 
+    sort_fields:
+        type: dict
```
julien-lang, line 246: Is this really a dict? From looking at the lines below, it looks more like a list<dict>, no? 
Or is it how tk-core handles this type of setting?

python/tk_multi_workfiles/browser_form.py
```diff
@@ -705,11 +709,183 @@ def _on_task_status_combo_changed(self, value):
...
+            return False
+
+        def disabled_filter(field):
+            # none of the fields are disabled
+            return False
+
+        # attach the filters
+        self._entity_field_menu.set_field_filter(field_filter)
+        self._entity_field_menu.set_checked_filter(checked_filter)
+        self._entity_field_menu.set_disabled_filter(disabled_filter)
```
julien-lang, line 782: IMO, this would be easier to read with `lambda` instead

```suggestion
        # attach the filters
        self._entity_field_menu.set_field_filter(
          # none of the fields are included
          lambda field: False
        )

        self._entity_field_menu.set_checked_filter(
          # none of the fields are checked
          lambda field: False
        )

        self._entity_field_menu.set_disabled_filter(
           # none of the fields are disabled
          lambda field: False
        )
```

python/tk_multi_workfiles/browser_form.py
```diff
@@ -705,11 +709,183 @@ def _on_task_status_combo_changed(self, value):
...
+                    fc, fsa, sort_actions
+                )
+            )
+
+        # Add actions to the entity Menu
+        self._entity_field_menu.add_group(sort_actions, "Sort menu")
+        # Remove the separator from the list
+        sort_actions.remove(separator)
+
+    def load_sort_data(self, field, sort_action, actions_list, **sort_order):
```
julien-lang, line 836: I don't understand this syntax. Do you mean you want `sort_order` to be a mandatory named parameter? If that's so, I believe the syntax is:

```suggestion
    def load_sort_data(self, field, sort_action, actions_list, *, sort_order):
```

What did I miss?

python/tk_multi_workfiles/browser_form.py
```diff
@@ -705,11 +709,183 @@ def _on_task_status_combo_changed(self, value):
...
             self.my_tasks_model.load_and_refresh()
         else:
-            self.my_tasks_model.load_and_refresh(
-                extra_filter=["sg_status_list", "is", value.lower()]
-            )
+            self._extra_filter = ["sg_status_list", "is", value.lower()]
+            self.my_tasks_model.load_and_refresh(extra_filter=self._extra_filter)
 
         self.my_tasks_model.data_refreshed.emit(True)
-        self.my_tasks_model.async_refresh()
```
julien-lang, line 715: Why do we remove this?

julien-lang (commented):
Wow it's odd to see a 4-digit ticket ID ;)

I have a few initial questions but mostly trying to understand how the code works.

info.yml
carlos-villavicencio-adsk: If I understand correctly, it is a dict on the shape:

`
{
  "tasks": [
    {"field_code": "due_date", "display_name": "Due Date"},
    ...
  ]
}
```

File: python/tk_multi_workfiles/browser_form.py (hunk 4 above)
Comment (carlos-villavicencio-adsk) [reply python/tk_multi_workfiles/browser_form.py:715]: It's already part of `load_and_refresh` implementation.

File: python/tk_multi_workfiles/browser_form.py (hunk 2 above)
Comment (carlos-villavicencio-adsk) [reply python/tk_multi_workfiles/browser_form.py:782]: Good call, it was directly copied from the previous ticket implementation.

File: python/tk_multi_workfiles/browser_form.py (hunk 3 above)
Comment (carlos-villavicencio-adsk) [reply python/tk_multi_workfiles/browser_form.py:836]: It shouldn't be mandatory, otherwise it will fail on line 826. 

The invocation actually looks like this: `sort_order="desc"`

I think it would be best if I rename this to `**kwargs` or just `sort_order="asc"`.

File: python/tk_multi_workfiles/browser_form.py (hunk 4 above)
Comment (julien-lang) [reply python/tk_multi_workfiles/browser_form.py:715]: So it was never needed?

File: info.yml (hunk 1 above)
Comment (julien-lang) [reply info.yml:246]: Reading this code: https://github.com/shotgunsoftware/tk-core/blob/v0.22.3/python/tank/platform/validation.py#L541 

I believe the type should be `list` not `dict`.

Am I wrong?

File: python/tk_multi_workfiles/browser_form.py (hunk 4 above)
Comment (carlos-villavicencio-adsk) [reply python/tk_multi_workfiles/browser_form.py:715]: [Exactly](https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/b4f8e76b59efb72afab39d317c42651d5793cbde/python/tk_multi_workfiles/entity_models/extended_model.py#L100).

File: info.yml (hunk 1 above)
Comment (carlos-villavicencio-adsk) [reply info.yml:246]: Can you elaborate more? I don't understand.

If you see [this line](https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/158/files#diff-947bc20553523554a0e6f03cd625faac913ccaaa738d7e004a9d6e3b401eaa6cR778), I see clearly that the setting is a `dict`, otherwise the `.get("something")` would fail.

File: info.yml [hunk 5]
```diff
@@ -242,6 +242,27 @@ configuration:
...
+        description: Controls what fields appear for sorting in the UI.
+                     Currently the sort menu only shows on the left tasks tab.
+                     You can add any field that would normally be suitable for sorting a query
+                     by for the tab's given entity type.
+        allows_empty: True
+        default_value:
+            tasks:
+                - { field_code: due_date, display_name: Due Date }
```
pscadding, line 254: One difference from the panel app I note is not including an[ optional `default` property](https://github.com/shotgunsoftware/tk-multi-shotgunpanel/blob/master/info.yml#L48C71-L48C78), to define which of the settings would be used by default. Was this intentional?

python/tk_multi_workfiles/browser_form.py
pscadding: It makes sense. I guess I would just be careful in case we were relying on a double refresh in someway. Could this introduce a race condition in someway?

python/tk_multi_workfiles/browser_form.py
```diff
@@ -705,11 +709,167 @@ def _on_task_status_combo_changed(self, value):
...
+        # Connect
+        self._my_tasks_form.sort_menu_button.clicked.connect(
+            lambda: self._entity_field_menu.exec_(QtGui.QCursor.pos())
+        )
+
+        # the fields manager is used to query which fields are supported
+        # for display. it can also be used to find out which fields are
+        # visible to the user and editable by the user. the fields manager
+        # needs time to initialize itself. once that's done, the widgets can
+        # begin to be populated.
```
pscadding, line 754:
```suggestion
        # The fields manager is used to query which fields are supported
        # for display. It can also be used to find out which fields are
        # visible to the user and editable by the user. The fields manager
        # needs time to initialize itself. Once that's done, the widgets can
        # begin to be populated.
```

python/tk_multi_workfiles/browser_form.py
```diff
@@ -705,11 +709,167 @@ def _on_task_status_combo_changed(self, value):
...
+
+    def _field_filters(self):
+        """
+        Define a few simple filter methods for use by the menu
+        """
+        # Attach the filters
+        # None of the fields are included, checked, and disabled at startup
+        self._entity_field_menu.set_field_filter(lambda field: False)
+        self._entity_field_menu.set_checked_filter(lambda field: False)
+        self._entity_field_menu.set_disabled_filter(lambda field: False)
```
pscadding, line 770: I'm not sure I understand what the fields manager and field filters are for?

Based on the comment it sounds like it will be filtering out any options that were configured if the user doesn't have visibility of the field in FPTR, is that correct?
I'm not sure I understand how this is working without a deeper dive.

Should the panel app be doing this as well?

pscadding (approved):
Looks good to me.

If I'm being picky, I don't like the sort button next to the new task button, that row of widgets looks messy. I might have put it next to the search box below maybe? Feel free to ignore though.

python/tk_multi_workfiles/browser_form.py
carlos-villavicencio-adsk: I didn't find any possible scenario for race conditions here. This line was introduced by me on #147 based on a customer contribution. However, I didn't have much context about the model back then.

info.yml
carlos-villavicencio-adsk: Yes, it was. This is a different factor from your original feature on the panel. For performance reasons, we initially display the default sorting that comes from the server, and then, the user can do the sorting.

python/tk_multi_workfiles/browser_form.py
carlos-villavicencio-adsk: This was implemented based on https://github.com/shotgunsoftware/tk-multi-shotgunpanel/pull/67. 

As far as I understand, the `field_manager` defined like this:
```py
        fields_manager = shotgun_fields.ShotgunFieldManager(
            self, bg_task_manager=self._task_manager
        )
```
 Brings all possible fields from the server. Later in this method it hides everything. Then, it starts displaying the fields based on the YAML settings. 

I'm not sure the main reason behind this TBH.

julien-lang (approved):

julien-lang (changes requested):
Could you add the "sort" icon please? Just like in the tk-multi-shotgunpanel?

image

Then, can we talk about improving the layout? Like Phil was saying the sort button aside from the "create a task" button is not the best.

info.yml
julien-lang: I would like to challenge the previous comment. The data type is `list<dict>` and not `dict<list>`. So the current definition of info.yml is wrong.

I think I found one similar setting that was properly implemented: https://github.com/shotgunsoftware/tk-multi-publish2/blob/v2.10.3/info.yml#L74

carlos-villavicencio-adsk:
@julien-lang @pscadding  After a discussion with our design consultant, we decided to go with this layout. PR is updated now.

image

julien-lang (approved):

pscadding (approved):
LGTM

eduardoChaucaGallegos (approved):

PR #180, opened 2026-05-11
SG-37743 Fix template resolution in the case of pipeline step rename

yungsiow:
**Precondition:**
- User creates a workfile under a pipeline step
- User renames the pipeline step in SG
- User runs 'tank unregister_folders' command for original step folder on disk
- User manually renames step folder on disk to match new step name

**Problem:**
- Workfiles under the renamed step would not show up in the File Open app
- Clicking New File from the app would fix the issue

**Solution:**
- Catch the exception when an initial attempt to resolve the template path fails (this happens due to the fact that the path cache is still storing the old step name)
- Mimic the behaviour triggered by New File action, which is to invoke sgtk.create_filesystem_structure() - this will re-query SG and create an updated path cache with the proper links
- The fix implicitly fixes the re-name issue which incurs a slight cost in an edge case situation, but for happy path, does not add any extra work

### DIFF
--- a/python/tk_multi_workfiles/file_finder.py
+++ b/python/tk_multi_workfiles/file_finder.py
@@ -544,9 +544,9 @@ class FileFinder(QtCore.QObject):
             except TankError as exc:
                 # In this case, we cannot continue with any file system resolution, so just exit early instead.
                 self._app.log_debug(
-                    "Could not resolve template fields for context %s after "
+                    f"Could not resolve template fields for context {context} after "
                     "reconciling filesystem structure; returning no work files. "
-                    "Last error: %s" % (context, exc)
+                    "Last error: {exc}"
                 )
                 return []
 
@@ -601,8 +601,7 @@ class FileFinder(QtCore.QObject):
             return True
         except Exception as e:
             self._app.log_warning(
-                "Could not reconcile filesystem structure for context %s: %s"
-                % (context, e)
+                f"Could not reconcile filesystem structure for context {context}: {e}"
             )
             return False
 
--- a/python/tk_multi_workfiles/file_finder.py
+++ b/python/tk_multi_workfiles/file_finder.py
@@ -571,7 +571,7 @@ class FileFinder(QtCore.QObject):
         )
         return work_file_paths
 
-    def _reconcile_filesystem_for_context(self, context):
+    def _reconcile_filesystem_for_context(self, context: sgtk.Context) -> bool:
         """
         Reconcile filesystem structure with current Shotgun state for the given
         context. Equivalent to the work `+ New File` triggers via
--- a/python/tk_multi_workfiles/file_finder.py
+++ b/python/tk_multi_workfiles/file_finder.py
@@ -529,12 +529,26 @@ class FileFinder(QtCore.QObject):
         work_fields = []
         try:
             work_fields = context.as_template_fields(work_template, validate=True)
-        except TankError as e:
-            # could not resolve fields from this context. This typically happens
+        except TankError as exc:
+            # Field resolution failed. This typically happens
             # when the context object does not have any corresponding objects on
-            # disk / in the path cache. In this case, we cannot continue with any
-            # file system resolution, so just exit early insted.
-            return []
+            # disk / in the path cache. It is also possible that the cause is a stale
+            # path cache — for example, after an entity short_name was renamed and
+            # `tank unregister_folders` ran without folders being recreated.
+            # In case it is the latter, mirror what "+ New File" does by reconciling the
+            # filesystem structure against current Shotgun state, then retry once.
+            if not self._reconcile_filesystem_for_context(context):
+                return []
+            try:
+                work_fields = context.as_template_fields(work_template, validate=True)
+            except TankError as exc:
+                # In this case, we cannot continue with any file system resolution, so just exit early instead.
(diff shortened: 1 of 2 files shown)


python/tk_multi_workfiles/file_finder.py
```diff
@@ -557,6 +571,41 @@ def _find_work_files(self, context, work_template, version_compare_ignore_fields
         )
         return work_file_paths
 
+    def _reconcile_filesystem_for_context(self, context):
+        """
+        Reconcile filesystem structure with current Shotgun state for the given
+        context. Equivalent to the work `+ New File` triggers via
+        FileAction.create_folders, but safe to call from a background worker
```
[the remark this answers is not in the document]
yungsiow: No thanks

python/tk_multi_workfiles/file_finder.py
```diff
@@ -557,6 +571,41 @@ def _find_work_files(self, context, work_template, version_compare_ignore_fields
...
+            self._app.sgtk.create_filesystem_structure(
+                ctx_entity.get("type"),
+                ctx_entity.get("id"),
+                engine=self._app.engine.instance_name,
+            )
+            return True
+        except Exception as e:
+            self._app.log_warning(
+                "Could not reconcile filesystem structure for context %s: %s"
+                % (context, e)
```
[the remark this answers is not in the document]
yungsiow: I like this suggestion, but am not sure if we are breaking any toolkit convention here by doing this.  @carlos-villavicencio-adsk would this be ok?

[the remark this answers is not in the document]
carlos-villavicencio-adsk: I'd prefer to stick to warning

python/tk_multi_workfiles/file_finder.py
```diff
@@ -529,12 +529,26 @@ def _find_work_files(self, context, work_template, version_compare_ignore_fields
         work_fields = []
         try:
             work_fields = context.as_template_fields(work_template, validate=True)
-        except TankError as e:
-            # could not resolve fields from this context. This typically happens
+        except TankError as exc:
```
julien-lang, line 532: Can we catch this specific case with a specific exception or code instead?

python/tk_multi_workfiles/file_finder.py
```diff
@@ -529,12 +529,26 @@ def _find_work_files(self, context, work_template, version_compare_ignore_fields
...
+            if not self._reconcile_filesystem_for_context(context):
+                return []
+            try:
+                work_fields = context.as_template_fields(work_template, validate=True)
+            except TankError as exc:
+                # In this case, we cannot continue with any file system resolution, so just exit early instead.
+                self._app.log_debug(
+                    "Could not resolve template fields for context %s after "
+                    "reconciling filesystem structure; returning no work files. "
+                    "Last error: %s" % (context, exc)
```
julien-lang: please use fstring or format.

python/tk_multi_workfiles/file_finder.py
```diff
@@ -557,6 +571,41 @@ def _find_work_files(self, context, work_template, version_compare_ignore_fields
...
+            return False
+
+        try:
+            self._app.sgtk.create_filesystem_structure(
+                ctx_entity.get("type"),
+                ctx_entity.get("id"),
+                engine=self._app.engine.instance_name,
+            )
+            return True
+        except Exception as e:
```
julien-lang, line 602: can we get more specific exception to catch?

python/tk_multi_workfiles/file_finder.py
```diff
@@ -557,6 +571,41 @@ def _find_work_files(self, context, work_template, version_compare_ignore_fields
...
+            self._app.log_warning(
+                "Could not reconcile filesystem structure for context %s: %s"
+                % (context, e)
```
julien-lang: fstring or format

julien-lang (changes requested):

python/tk_multi_workfiles/file_finder.py
yungsiow: Hi Julien, the original behaviour caught TankError here.  Are you wanting to change the original behaviour within this PR?

python/tk_multi_workfiles/file_finder.py
yungsiow: So, I think in this case, we really do want a broad Exception catch here.  Reason being that the attempt to "repair" the situation by calling create_filesystem_structure() could potentially lead to any number of exceptions (file system interactions can be unpredictable in many ways).  On the receiving end, we really want to make sure that this attempt at repair cannot become a risk for an unhandled point of failure (which would make it a bad fallback mechanism).

Claude suggested that a better approach would be to use log_exception() instead of log_warning() here to show the full traceback and improve our understanding of the full context of a caught exception for better debugging in this situation.  I agree with this suggestion as a compromise.  Let me know what you think!

python/tk_multi_workfiles/file_finder.py
yungsiow: fixed.

python/tk_multi_workfiles/file_finder.py
yungsiow: fixed.

julien-lang (approved):

Issue #183, opened 2026-07-17
Open Previous Version does nothing with PySide6


KimGoddardJr:
https://github.com/shotgunsoftware/tk-multi-workfiles2/blob/90290e031ffe1c0d802c3c3a281f7b468506f6c4/python/tk_multi_workfiles/file_open_form.py#L235

PySide6 sends a checked boolean when the menu action is clicked. The lambda treats that boolean as the file action, so _perform_action(False) silently does nothing. Putting checked first preserves the real action and opens the selected version.

```python
lambda checked=False, a=action: self._perform_action(a)
```

julien-lang:
Hi @KimGoddardJr the issue should now be fixed in [v0.18.0](https://github.com/shotgunsoftware/tk-multi-workfiles2/releases/tag/v0.18.0).

PR #184, opened 2026-07-21
SG-44704 Fix Workfiles2 context-menu actions under PySide6 6.11

dzohlen:
I am currently writing a custom engine which required a newer version of PySide6 (6.7.x+) - I am using 6.11.1 to be precise.
The app works just fine as usual. When trying to use any context-menu action in tk-multi-workfiles2, however, it would just silently fail.

I was able to narrow down the issue and solve the problem with this PR on my end. As far as I can tell, both new and older versions of PySide6 are working as intended with this patch.

Others have also encountered this issue, I believe. There is one other open PR and one open issue that seem related:
- https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/85
- https://github.com/shotgunsoftware/tk-multi-workfiles2/issues/183

Below you can find a more detailed explanation of what I believe the issue to be and what the actual fix was for me.

## Issue

Workfiles2 currently connects context-menu actions using:

```python
q_action.triggered[()].connect(
    lambda a=action, checked=False: self._perform_action(a)
)
```

Under PySide6 6.11.1, `triggered[()]` invokes the callback with the signal’s `checked` boolean argument.

Python assigns that positional value to the lambda’s first parameter, `a`. As a result, the default value `a=action` is not used, and `a` becomes `False`.

This causes Workfiles2 to call:

```python
self._perform_action(False)
```

The method then exits without executing the selected context-menu action.

Older PySide6 versions tested (i.e. 6.5.8 in my case), do not pass the `checked` argument when using `triggered[()]`, which is why the existing implementation works in those environments.

## Fix

Connect to the signal directly and accept the `checked` argument as the first callback parameter:

```python
q_action.triggered.connect(
    lambda checked=False, a=action: self._perform_action(a)
)
```

This ensures that the signal’s boolean value is assigned to `checked`, while the Workfiles2 action remains captured by `a`.

## Testing

Verified the behavior with:

* PySide6 6.5.8
* PySide6 6.11.1

The updated callback preserves the correct Workfiles2 action and allows context-menu actions to execute across all tested versions.

### DIFF
--- a/python/tk_multi_workfiles/file_open_form.py
+++ b/python/tk_multi_workfiles/file_open_form.py
@@ -231,8 +231,8 @@ class FileOpenForm(FileFormBase):
                 add_separators = True
             else:
                 q_action = QtGui.QAction(action.label, menu)
-                q_action.triggered[()].connect(
-                    lambda a=action, checked=False: self._perform_action(a)
+                q_action.triggered.connect(
+                    lambda checked=False, a=action: self._perform_action(a)
                 )
                 menu.addAction(q_action)
                 add_separators = True


julien-lang:
Thanks for your contribution @dzohlen. We are going to review and test this. And we will be happy to merge if we see no regression.

Besides that, you don't have any problem using Toolkit with Qt 6.11?

Closes #85, #183

dzohlen:
I'm happy to hear that, @julien-lang. I hope everything passes the review and tests.

I haven't encountered any other issues with Qt 6.11 so far. However, we are currently primarily testing PySide6 6.11.1 in the context of our newly developed engine due to version restrictions. For now, I would say I'm cautiously optimistic.

julien-lang (approved):
LGTM!
I confirm this change is compatible with PySide 2/5.15 and PySide6/6.5.

julien-lang:
Released in [v0.18.0](https://github.com/shotgunsoftware/tk-multi-workfiles2/releases/tag/v0.18.0).

PR #185, opened 2026-07-21
Fix python3 migration bug

JacobTerritory:


julien-lang:
Hi @JacobTerritory, thanks for your contribution.

Could you provide more information about the problems you are trying to fix here, please?
We are not aware of any Python 3 related issues with tk-multi-workfiles2.

JacobTerritory:
Apologies, we forked the repository to add some additional features which I didn't mean to push back. 

However, the latest commit here, to `interactive_open_action.py` is a bugfix due to a change from Python 2.7 -> 3.x. 

In python 2.7,  NoneType < Int would evaluate to True, but in python 3.x it raises a TypeError. So we must check for NoneType before evaluating the if condition.

julien-lang:
Hi @JacobTerritory, thanks for answering my questions. I have a couple of follow up questions.

Regarding the additional feature, would you be able to remove them from this PR, please?

About the Python 3 regression, do you have a way to reproduce this issue? I mean, I agree with the logic and your explaination. But it would be easier for us to have clear reproduction steps so we can mark this as a regression and prioritize it.

Then, last point, the CI reports issues with code styling. Could you run pre-commit locally to fix it please and then push the changes?

Thanks 🙏

JacobTerritory:
I'm not sure on exact reliable reproduction steps, because this is obviously not a common issue, otherwise it would have been flagged years ago. 

For us, it's happening whenever an artist tried opening a scene file at the project level, from another artists user folder other than their own.

JacobTerritory:
I didn't want to modify this branch on our fork, so I've instead created a new branch, and dropped the other 2 commits. 

I've submitted another PR here. (I don't think I have permission to change the branch on this PR)
https://github.com/shotgunsoftware/tk-multi-workfiles2/pull/186

julien-lang:
Thanks @JacobTerritory, can we close this PR then?

PR #186, opened 2026-08-28
Fixed remnant bug from changed behavior in python 2 where int < None …

JacobTerritory:
…would evaluate to False rather than a ValueError. Manifested when openning another user's user folders

### DIFF
--- a/python/tk_multi_workfiles/actions/interactive_open_action.py
+++ b/python/tk_multi_workfiles/actions/interactive_open_action.py
@@ -55,7 +55,7 @@ class InteractiveOpenAction(OpenFileAction):
             and (not self._workfiles_visible or not self.file.is_local)
         ):
             # opening a publish and either not showing work files or the file isn't local
-            if self.file.version < max_publish_version:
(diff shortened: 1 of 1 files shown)


python/tk_multi_workfiles/actions/interactive_open_action.py
```diff
@@ -52,7 +55,7 @@ def execute(self, parent_ui):
             and (not self._workfiles_visible or not self.file.is_local)
         ):
             # opening a publish and either not showing work files or the file isn't local
-            if self.file.version < max_publish_version:
+            if max_publish_version and self.file.version < max_publish_version:
```
julien-lang, line 58: I understand that the change of behaviour between Python 2 and 3 with `Int<None` and I agree this fix makes sense.

However, I am unable to find a way to reproduce this problem.

I mean, if `max_publish_version` is `None`, that means there are no `file_versions` with `is_published`. In this situation, the condition on L54 (`and self.file.is_published`) will never pass so we never get to this line.

Would it be possible to provide reproduction steps for this issue, please?

PR #187, opened 2026-09-11
SG-44537 Remove monitor_qobject_lifetime debug QObject tracking

stevelittlefish:
## Summary

Fixes the Maya 2027.2 crash-on-save reported in SG-44537: after importing a scene / creating a new file, saving via Workfiles2 would hang and then crash Maya (Windows Event Viewer: `0xc0000374` STATUS_HEAP_CORRUPTION). Regression vs Maya 2027.0; not reproducible on Maya 2026.

## Root cause

`monitor_qobject_lifetime()` in `util.py` is a **debug-only** diagnostic that logs when a QObject is destroyed by connecting a Python slot to each tracked object's `destroyed` signal:

```python
obj.destroyed.connect(lambda m=msg, u=uid: _on_qobject_destroyed(m, u))
```

It was wired onto ~9 objects across the Workfiles UI, including the My Tasks item delegate (`QStyledItemDelegate`). Under the Qt/PySide6 build shipped in Maya 2027.2, activating a Python slot from inside a QObject's `~QObject()` causes Qt to take a reference to the object while it is being destroyed — which the debug crash stack shows aborting in `QtSharedPointer::ExternalRefCountData::getAndRef` (and, with the assert compiled out in release, corrupting the heap). This matches both the debug-build assert and the release heap-corruption dumps in the ticket, and explains why minimal `cmds.workspace()` repro scripts never crashed (no tracked QObjects / delegates to activate mid-destruction).

## Fix

Remove the `monitor_qobject_lifetime` debug facility entirely — it is a pure diagnostic with no production value:

- Delete `monitor_qobject_lifetime`, `_on_qobject_destroyed`, `report_non_destroyed_qobjects`, and `_g_monitored_qobjects` from `util.py`.
- Remove all call sites and their imports (`browser_form.py`, `file_finder.py`, `entity_tree/entity_tree_form.py`, `file_form_base.py`, `my_tasks/my_tasks_form.py`, and the `report_non_destroyed_qobjects()` call in `work_files.py`'s `dbg_info`).

No behaviour change beyond removing the debug logging.

## Testing

- Verified manually in **Maya 2027.2** with this branch loaded live via a `dev` descriptor (confirmed through `app.disk_location`). The original repro (open/close Workfiles dialog, import → save) no longer hangs or crashes.
- `black` and `flake8` pass on all changed files.

## Notes

- Heap corruption is timing-dependent, so this is validated by repeated manual repro rather than an automated test.
- A **separate** crash on Maya shutdown (`0xc0000005` access violation, distinct from this heap-corruption bug) was observed during investigation and will be filed as its own ticket — it is **not** addressed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[SG-44537]: https://autodesk.atlassian.net/browse/SG-44537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


carlos-villavicencio-adsk (approved):
LGTM. We need to do an extensive QA on all DCCs for this one

